PR #16, opened 2023-07-13
Add a path based rounded rectangle

atulpatare:
This is an alternative implementation of a rounded rectangle.
Working:
 - The rounded edge is rendered by using a set of points that defined the edge.
 
 Will continue to optimise if some better way to do this exists!

This is how it looks
rounded-rect


AP-Atul:
I dont think the existing shader works correctly. Does it?
I'm just learning shader from "book of shader" and "lumagl"

atulpatare:
Got this from the slack https://codepen.io/grahambates/pen/bGQLwGJ?editors=0011
Here the rounded rect does work, but with the graph layer I wan't able to use it. Even tried updating the stylesheet.

dylans:
I'm going to land this to make it easier to try/experiment with.

Issue #15, opened 2023-07-11
How do I run examples


atulpatare:
Im not able to run the examples. Steps I performed

```
git clone <repo>
npm i
npm run build # build works
```

I am not able to run the tests, getting below error

```
TSError: ⨯ Unable to compile TypeScript:
error TS6046: Argument for '--module' option must be: 'none', 'commonjs', 'amd', 'system', 'umd', 'es6', 'es2015', 'es2020', 'es2022', 'esnext', 'node12', 'nodenext'.
```

I am using 
node `v18.16.1` and tried with `v16`

If you can you update the Readme or add steps to run different modules, examples and the website that will be great. Thanks

For testing the graph layer, I just copied required jsx files into my project and tested it out.

donmccurdy:
Work in progress:

- https://github.com/visgl/deck.gl-community/pull/31

donmccurdy:
Adding documentation for running examples in:

- https://github.com/visgl/deck.gl-community/pull/47

PR #31, opened 2024-02-01
chore(deps): Upgrade to ocular-dev-tools@v2.0.0-alpha.22

donmccurdy:
As described in #15, examples don't seem to be working on the master branch, and I wasn't able to run the test suite either, so perhaps there's been some bitrot over time. As a first step toward updates, I've upgraded to the latest ocular-dev-tools and made downstream changes:

- Builds
  - All modules build, output looks valid
  - Most modules (all except 'layers') build CJS and ESM modules
  - 'layers' module only builds ESM. It's the only module here that's TypeScript... could be related? Can't find that in this repo's config anywhere.
- Examples
  - Moved all from Webpack to Vite
  - 'minimal' example builds and runs
  - 'bing-maps' and 'react-graph-gl' examples resolve workspace dependencies correctly, but fail with errors that suggest incorrect versions of loaders-gl dependencies.
- Moved 'modules-wip/submodule' to 'modules/template' to simplify builds
- Fixed or added basic tests for three modules, the other two are failing. I wasn't able to get ocular-dev-tools running the Jest test suite properly — we need [special handling for ESM](https://jestjs.io/docs/ecmascript-modules) — so I'm just letting Lerna execute the tests for the moment.
- Consolidate most devDependencies at the project root workspace

Next steps:

1. Lock in versions of deck.gl and loaders.gl to get other examples working
2. Enable CI
3. Fix remaining tests
    - Fix Jest/Ocular issues?
    - Switch to `tape` to match Luma and Deck repositories?
    - Some issues likely fixed by deck.gl and loaders.gl upgrades


examples/minimal/package.json
```diff
@@ -1,21 +1,15 @@
...
   },
   "devDependencies": {
-    "@babel/preset-env": "^7.20.2",
-    "@babel/preset-react": "^7.0.0",
-    "babel-loader": "^9.1.2",
-    "html-webpack-plugin": "^5.5.0",
-    "webpack": "^5.75.0",
-    "webpack-cli": "^5.0.1",
-    "webpack-dev-server": "^4.11.1"
+    "vite": "^5.0.12"
```
ibgreen, line 13: Don't we usually also need to install typescript?

examples/react-graph-gl/index.jsx
```diff
@@ -1,8 +1,9 @@
 import React, {Component} from 'react';
 
-import GraphGL, {D3ForceLayout, JSONLoader, NODE_TYPE} from 'modules/react-graph-layers';
+import {D3ForceLayout, JSONLoader, NODE_TYPE} from 'deck-graph-layers';
+import {GraphGL} from 'react-deck-graph-layers';
 
-import SAMPLE_GRAPH_DATASETS from '../../utils/data/sample-datasets';
+import SAMPLE_GRAPH_DATASETS from 'react-deck-graph-layers/test/utils/data/sample-datasets.js';
```
ibgreen, line 6: Perhaps an underscore export for these in the `'react-deck-graph-layers` module? Raking for files like this seems fragile.

examples/react-graph-gl/index.jsx
```diff
@@ -1,8 +1,9 @@
 import React, {Component} from 'react';
 
-import GraphGL, {D3ForceLayout, JSONLoader, NODE_TYPE} from 'modules/react-graph-layers';
+import {D3ForceLayout, JSONLoader, NODE_TYPE} from 'deck-graph-layers';
```
ibgreen, line 3: Nit: I would probably rename files to `.tsx` while we are at it, and go all typescript, a JS/TS mix is just confusing in the long run...

ibgreen:
>  Fix Jest/Ocular issues?
> Switch to tape to match Luma and Deck repositories?

- I would very much like to move all the vis.gl repos to "jest compatible" tests. 
- Ultimately we likely want to use a faster jest-compatible test runner like vitest or bun. 
- It would be great if it was possible to add support for "jest style tests" in ocular.
- I don't have a good migration strategy in mind yet, open to suggestions. Perhaps the ocular `test` script could call `test-tape` and `test-jest` and frameworks could start opting in to jest without having to port all tests.

examples/minimal/package.json
donmccurdy: Shouldn't be required in this example, but at the project root I think you're right — I'll add it there.

examples/react-graph-gl/index.jsx
donmccurdy: Yeah, I'd prefer to avoid the deep import too. Worried about adding an export though, unsure what bundlers will do if the imported data is missing from the npm package, and probably we don't want it published there. 

Maybe a new module `modules/test-utils` with shared data and utils for tests? Or perhaps a shared alias in a config file?

examples/minimal/package.json
donmccurdy: Or if we're converting examples to TS then we should add it here too, yes!

examples/react-graph-gl/index.jsx
donmccurdy: Going to hold this for a followup PR if that's OK! But agreed, and maybe implied by #14.

examples/react-graph-gl/index.jsx
ibgreen: Yes... FWIW I personally think all examples / all code in our repos should be `.ts/.tsx`. If we do that and want users to be able to copy these folders and run on their own, we may need the dependency in each example folder. That is how I do it in luma, loaders etc.

donmccurdy:
> Ultimately we likely want to use a faster jest-compatible test runner like vitest or bun.

Whoa! 🤯 I had no idea Bun and Vitest were Jest-compatible. That is helpful.

My expectation is that ESM changes in Deck and Luma v9 are going to affect how we run tests substantially. Jest marks their ESM support as "experimenal", which is frustrating, but with Bun and Vitest available I'm sure can manage. It might be best to land on a testing strategy for one of our ESM-ready repositories first before migrating the tests here, this repo still depends on Deck v8 for the time being.

ibgreen:
> It might be best to land on a testing strategy for one of our ESM-ready repositories first before migrating the tests here, this repo still depends on Deck v8 for the time being.

The counter arguments are 
- that this is a great repo for experimentation 
- two of the modules (both graph-layers and the planned editable-layers / nebula.gl module are already using jest, so we need to solve it here.
- we do need to get this module up to deck.gl v9 as well.

examples/react-graph-gl/index.jsx
donmccurdy: I've mode the dataset to `@deck.gl-community/test-utils` for now, and switched that and the template module to TypeScript + Vitest. I'm not sure if "may need the dependency in each example folder" means an npm dependency or just copying the data? For examples and tests I think having it be very obvious what the data looks like can be helpful, maybe an inline dataset would eventually be the choice here, but that's for another PR I think.

donmccurdy:
I've switched the `template` and `test-utils` (new) modules to Vitest for experimentation purposes. So far that seems much easier to work with than Jest, and ESM+TS come out of the box.

examples/react-graph-gl/index.jsx
ibgreen: Sounds good. I would vote for having the datasets for the examples in the example folder. They should be standalone as far as possible

Having datasets from all different modules in the test-utils seems to centralized.

PR #32, opened 2024-02-02
chore(ci): Add CI

donmccurdy:
Copied from deck.gl's `test.yaml`, with modifications. Currently `yarn test` just runs a linter, but this will run unit tests as well after #31.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,34 @@
+name: test
+
+# On every pull request, but only on push to master
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  test-node:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Use Node.js
+        uses: actions/setup-node@8f152de45cc393bb48ce5d89d36b731f54556e65 # v4.0.0
+        with:
+          node-version: '18.x'
+
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  test-node:
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
ibgreen: These commit SHAs seem odd, why can't we use versions?

ibgreen (approved):

.github/workflows/test.yaml
donmccurdy: Looks like this came into Deck with https://github.com/visgl/deck.gl/pull/7887, with context in https://github.com/visgl/deck.gl/issues/7868, which is where I've copied this from. I don't feel strongly about the syntax. I do use these SHAs in my own repositories, but only because I have it automated, see...

https://github.com/donmccurdy/glTF-Transform/pull/1129

... I probably would choose version numbers as well if maintaining these by hand, unless it was a security-sensitive action.

ibgreen: OK, thanks for clarifying. If the syntax of this file allows for comments, I would mention that we use SHAs for security reasons, otherwise it can look rather arbitrary and hacky to maintainers.

donmccurdy: I read a little more on this, and GitHub's own advice is to do this only for untrusted third-party actions. Both of these are core actions provided by GitHub... perhaps for now I'll just revert to version numbers here.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

PR #37, opened 2024-03-28
chore: Add nebula.gl modules

ibgreen:
This PR forks nebula.gl into this repo withe the following module mapping:
| nebula module (path) | Description | Forked module |
| --- | --- | --- | 
| `nebula.gl`  (modules/main) | The core module |=> `deck.gl-editable-layers` |
| `@nebula.gl/edit-modes`  (modules/edit-modes)  | Optional edit modes| => `deck.gl-editable-layers` |
| `@nebula.gl/layers` (modules/layers) | The actual layers | => `deck.gl-editable-layers` 
| `@nebula.gl/editor` (modules/editor) | React wrapper | => `react-deck.gl-editable-layers` 
| `@nebula.gl/overlays` (modules/overlays) | React General  overlays | => `react-deck.gl-editable-layers` | 
| `react-map-gl-draw` (modules/react-map-gl-draw) | Non-deck-wrapper | => NOT FORKED |


chrisgervang:
This module grouping makes sense to me. I can imagine most of the features in `react-deck.gl-editable-layers` being implemented as a deck.gl widget, which would work for both react and pure-js applications.

ibgreen:
Yes I think the goal should be to sunset / remove the react module if at all possible. Whatever it offers should be available to all bindings.
- We should create a list of reasons why it is needed and brainstorm about ways we can solve those needs in react-independent ways.

felixpalmer (approved):

PR #39, opened 2024-04-04
chore(bing-maps): Update to TypeScript, move example to deckv9

donmccurdy:
Changes:

- move bing-maps module to TypeScript
- move bing-maps example to TypeScript and deck v9

Notes:

- eslint is a bit strict for kicking off a TS migration, I've loosened the eslint config to allow 'any' at least during the migration.
- using vitest here as a path of least resistance – that can be revisited, but I think the first priority is to get everything working on deck v9 and TypeScript first.
- tooltips currently display under the layer; will eventually need a workaround for that


examples/bing-maps/README.md
```diff
@@ -7,7 +7,7 @@ This is an example integrating deck.gl with Bing Maps using the deck.gl-bing-map
 To run this example, you need a [Bing Maps API key](https://www.microsoft.com/en-us/maps/create-a-bing-maps-key). You can either set an environment variable:
 
 ```bash
-export BingMapsAPIKey=<bing_maps_api_key>
+export VITE_BING_MAPS_API_KEY=<bing_maps_api_key>
```
ibgreen, line 10: Why VITE_ (search replace casualty or intentional?)

ibgreen (approved):
LGTM

examples/bing-maps/README.md
donmccurdy: Intended, but not a strong preference. Vite will automatically load env variables prefixed with VITE_, and ignores those that aren't, so our examples can be more standalone by using this convention:

https://vitejs.dev/guide/env-and-mode

Otherwise I believe the examples may be implicitly depending on the top-level config here:

https://github.com/visgl/deck.gl-community/blob/9be4ebc9cde1d7d0dc85368085439340a897600f/examples/vite.config.local.mjs#L27-L32

PR #43, opened 2024-04-05
chore(graph-layers): Migrate to TypeScript

donmccurdy:
Code builds in TypeScript, and tests pass against deck v9. Delaying working on the examples quite yet.

- https://github.com/visgl/deck.gl-community/issues/14


ibgreen:
Adding @dylans who initially ported these to deck.gl-community.

ibgreen (approved):
Thanks, it will be great if we can convert all code in the repo to typescript!

PR #44, opened 2024-04-05
chore(editable-layers): Migrate to Deck v9, fix an example

donmccurdy:
I believe `@deck.gl-community/editable-layers` may be in good enough shape to start testing an alpha release, if we've decided on the name.

Status:

- code builds and tests pass against deck v9
- some examples ported and working, others I haven't gotten to yet
- lots of miscellaneous eslint and typescript warnings


ibgreen:
> ... which makes me think there may be multiple versions of React coming from old dependencies or a CJS/ESM issue, but I haven't tracked down the cause yet.

Hmm, can this be resolved before bumping deck? Or is deck.gl/react pulling in a later version?

donmccurdy:
It's partly a problem of, the new packages have not been published yet, so we need to run in `yarn start-local` mode, but the dependencies at the workspace root are shared, and it's easy to get conflicts there ... but yes I think that's the right idea, just keep updating one piece at a time until it's all caught up. :)

donmccurdy:
Resolved the issue above — see changes in `examples/vite.config.local.mjs`.

ibgreen (approved):
Nit: maybe the changes to the modules could have been separated from the examples, this is rather overwhelming PR to review. 

But then again, we do need to get things working asap, so approved.

donmccurdy:
Thanks - happy to split a PR when helpful, too, that's no trouble. :)

PR #48, opened 2024-04-11
Overlays example

weberjavi:
This PR updates the overlays example using typescript and deckg v9. 
The example is quite minimal. It just uses a `HtmlOverlay` component with some `HtmlOverlayItems`.
Probably some more examples would be useful (`HtmlClusterOverlay` extended layer, overlay with interactive content...)


examples/editor-core/overlays/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+};
+export const INITIAL_VIEW_STATE = {
+  longitude: INITIAL_COORDS.lon,
+  latitude: INITIAL_COORDS.lat,
+  zoom: 17,
+  pitch: 0,
+  bearing: 0
+};
+export const MAPBOX_ACCESS_TOKEN =
+  'pk.eyJ1IjoiZ2Vvcmdpb3MtdWJlciIsImEiOiJjanZidTZzczAwajMxNGVwOGZrd2E5NG90In0.gdsRu_UeU_uPi9IulBruXA';
```
donmccurdy: I suspect that we probably should not include a token in the example. For me, locally, the example renders correctly without an access token as well. @ibgreen any preference between:

- omitting the token from the example
- including a (possibly undefined) `import.meta.env.VITE_MAPBOX_ACCESS_TOKEN` 
- ... something else?

I suppose this would affect how the website is deployed later on.

examples/editor-core/overlays/example.tsx
```diff
@@ -0,0 +1,86 @@
+import React, {useEffect, useState} from 'react';
+import DeckGL from '@deck.gl/react';
+import StaticMap from 'react-map-gl';
+import {INITIAL_COORDS, INITIAL_VIEW_STATE, MAPBOX_ACCESS_TOKEN} from './constants';
+import {HtmlOverlay, HtmlOverlayItem} from '@deck.gl-community/react-editable-layers';
```
donmccurdy, line 5: It looks like these exports are missing from the react-editable-layers `index.js` entrypoint, does that need to be included in this PR as well?

donmccurdy (commented):
Thank you @weberjavi!

examples/editor-core/overlays/constants.ts
ibgreen: We should definitely avoid tokens in the repos. Is the choice of mapbox here important? Can we just migrate to token-less basemaps, such as the CARTO basemaps? If not let's use an env variable.

examples/editor-core/overlays/example.tsx
weberjavi: My bad, I did not push those changes :/

weberjavi: done

examples/editor-core/overlays/constants.ts
weberjavi: I've downgraded react-map-gl to version 5 to avoid the need of using a token. Maybe this is convenient in other examples too

examples/editor-core/overlays/app.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
```
ibgreen, line 1: Nit: We should change the directory name from editor-core to editable-layers

ibgreen (approved):

examples/editor-core/overlays/app.tsx
ibgreen: Unrelated comment: I recommend creating a `modules/react` `@deck.gl-community/react` and move all general react components there. These overlays have no dependencies on editable-layers, and the same is true for a number of the components in react-graph-layers.

ibgreen: We want to eventually sunset the graph and editor specific react modules if at all possible, so stripping out common stuff makes a ton of sense.

examples/editor-core/overlays/constants.ts
donmccurdy: It was working without a token for me here, I'm not sure what's going on! In any case yes, we can change up the basemaps for any and all examples in a future PR. Thanks!

examples/editor-core/overlays/app.tsx
donmccurdy: @ibgreen What would you think of raising the nested examples up a level, so we have something like:

- examples/template
- examples/graph-layers
- examples/editable-layers-basic
- examples/editable-layers-advanced
- examples/editable-layers-overlays
- ...

It doesn't seem like deck.gl subdivides examples by submodule, we may want other folder divisions for the website, and I think it's slightly easier for users to explore a flat list.

@weberjavi I'll merge the PR after CI passes, no need to address any of the above in this PR. Thanks again!

donmccurdy (approved):

PR #49, opened 2024-04-11
chore: directory cleanup

ibgreen:
Reorganize docs files, delete unused code


modules/react/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
-  "name": "@nebula.gl/overlays",
-  "description": "React overlays for deck.gl",
+  "name": "@deck.gl-community/react",
+  "description": "React components for deck.gl",
   "license": "MIT",
   "version": "1.0.4",
   "author": "Georgios Karnas <georgios@uber.com>",
```
donmccurdy, line 7: Just flagging so we don't forget, these will need to be updated at some point:

- author
- version (optional)
- main / module / exports / types

modules/react/tsconfig.json
donmccurdy, line 1: Note that the `tsconfig.json` from another module might be a better template here.

donmccurdy (approved):

modules/react/package.json
ibgreen: OK, let's land and fix.

PR #51, opened 2024-04-12
chore(graph-layers): Cleanup code

ibgreen:
- Drop default exports
- Clean up file naming, remove nested index.js 
- Move docs to non-react module.
- etc


donmccurdy (approved):
Just curious, is the intention to avoid default exports in this repository in general?

ibgreen:
> Just curious, is the intention to avoid default exports in this repository in general?

Yes, I actively avoid them, they cause aliasing errors by letting the user mistakenly rename things, make it harder to search the codebase due to same thing being called different things, hide export errors, and also are not ideal for ESM.

donmccurdy:
Ok, thanks! I agree with your summary. 👍🏻

PR #52, opened 2024-04-12
editable-layers: prepare for publish

felixpalmer:
Update `package.json` to be public and correct version.

After doing this I ran `yarn build` at the root and then `npm publish` in `modules/editable-layers`. However npm is returning me an error:

```
npm ERR! code E402
npm ERR! 402 Payment Required - PUT https://registry.npmjs.org/@deck.gl-community%2feditable-layers - You must sign up for private packages
```

Any ideas @ibgreen @donmccurdy? `modules/layers` was published fine a month ago


modules/editable-layers/package.json
```diff
@@ -1,9 +1,8 @@
 {
-  "private": true,
```
donmccurdy, line 2: You'll also need to add:

publishConfig: {access: "public"}

Without it, npm assumes a scoped package is private by default the first time it's published, and requires payment. After that the publishConfig can be removed.

donmccurdy (approved):

modules/editable-layers/package.json
felixpalmer: Thanks, that got me one step closer - but now I have another cryptic error:

```
npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@deck.gl-community%2feditable-layers - Not found
npm ERR! 404
npm ERR! 404  '@deck.gl-community/editable-layers@9.0.0-alpha.1' is not in this registry.
```

I know it's not in the registry, that's why I'm trying to publish! 🙄

felixpalmer: `npm publish --access public` also no help, following https://docs.npmjs.com/creating-and-publishing-scoped-public-packages#publishing-scoped-public-packages

donmccurdy: That error is new to me, but looks a lot like:

https://stackoverflow.com/q/39115101

Possibly you don't have have permission to publish to the `@deck.gl-community` scope and it's just giving an opaque error message, @ibgreen would you be able to check the npm permissions?

ibgreen (approved):

donmccurdy:
As FYI — it looks like the build system is not outputting `.d.ts` files under the current configuration, which would prevent a successful publish. I'm working on that in:

- https://github.com/visgl/deck.gl-community/pull/55

felixpalmer:
Published! https://www.npmjs.com/package/@deck.gl-community/editable-layers

PR #53, opened 2024-04-12
chore(editable-layers): Fixes for type errors

donmccurdy:


modules/editable-layers/src/mode-handlers/draw-circle-by-bounding-box-handler.ts
```diff
@@ -29,7 +29,7 @@ export class DrawCircleByBoundingBoxHandler extends TwoClickPolygonHandler {
     }
 
     const firstClickedPoint = clickSequence[0];
-    const centerCoordinates = getIntermediatePosition(firstClickedPoint, event.groundCoords);
+    const centerCoordinates = getIntermediatePosition(firstClickedPoint, event.mapCoords);
```
ibgreen, line 32: Maybe just `coords` if we are changing? I know there was work on making nebula work with non-geospatial editing, but not sure where that stands.

ibgreen (approved):
Very nice

modules/editable-layers/src/mode-handlers/draw-circle-by-bounding-box-handler.ts
donmccurdy: See: 

https://github.com/visgl/deck.gl-community/blob/4878c854fd113c16429131d4dcb0fce374f28b2b/modules/editable-layers/src/edit-modes/types.ts#L32-L37

I think we need a term clearly distinct from screenCoords. Currently the project uses both groundCoords and mapCoords — I believe interchangeably? — and the type checker is not happy about that. Maybe worldCoords?

ibgreen: OK, let's now overthink it. mapCoords is probably fine for now.

PR #55, opened 2024-04-12
chore(build): Update builds to match deckgl

donmccurdy:
Work in progress.

Currently the build is not generating type declarations, so I'm attempting to reconfigure the build system to match deckgl. For editable-layers it appears to be working. The rest of the project has many tsc errors preventing a complete build under this setup.

~~Note that this build also omits CJS builds, but one thing at a time.~~


ibgreen (approved):

donmccurdy:
tsc, build, and tests passing with esbuild! esm, cjs, and d.ts output looks valid. Currently editable-layers will pass in strict mode but the rest doesn't, I've disabled that test for now so we can do this incrementally.

PR #58, opened 2024-04-15
chore(tests): Fix inconsistent test naming

donmccurdy:
Renames all tests from `*.test.ts` to `*.spec.ts`.


ibgreen (approved):
All other vis.gl repos consistently use `.spec.ts` so I would stay with that for now.

donmccurdy:
Sounds good – switched all tests to .spec.ts!

PR #59, opened 2024-04-15
chore(graph-layers): Restore pack-images script

donmccurdy:
The graph-layers module uses a script to generate the marker image spritesheet, which needed fixes since the repo has been migrated to TypeScript. I've also fixed the ESLint and TSC warnings for the script in the process. The get-pixels / save-pixels packages have been unmaintained for a few years now, so I've replaced both with [`ndarray-pixels`](https://github.com/donmccurdy/ndarray-pixels). (Disclaimer: I wrote 'ndarray-pixels' based on a combination of the older two).

As a side effect of using async/await, markers in the sprite sheet are now sorted alphabetically.


ibgreen (approved):
Neat. This should almost be its own module, rather than hidden in graph-layers...

donmccurdy:
Merging for now, but maybe the top-level `scripts/` folder? Similar to what deck.gl appears to be doing. I think the expectation then would be any script dependencies are added to the workspace as devDependencies.

ibgreen:
I was thinking that a `MarkerLayer` together with this type tooling is much more generic than a graph-layers module. 

There are probably also layers in editable-layers module that are quite generic. 

They could for instance be moved to the layers module.

PR #63, opened 2024-04-16
chore(examples): Add new editable-layers dependency

donmccurdy:
Changes: 

- Include `@deck.gl-community/editable-layers` dependency in examples, now that it has been published as an alpha release
- Clean up old references to `@deck.gl-community/bing-maps` in examples and website, replaced with `deck.gl-bing-maps`


tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "skipLibCheck": true,
     "baseUrl": ".",
     "paths": {
-      "@deck.gl-community/bing-maps": ["modules/bing-maps/src"],
```
felixpalmer: This module has never been published. Would it not be better to use `@deck.gl-community/bing-maps` and update `modules/bing-maps/package.json`?

felixpalmer (changes requested):

tsconfig.json
ibgreen: I didn't realize that. If it has never been published, my vote is that we go with the community namespace.

ibgreen (approved):

tsconfig.json
donmccurdy: Done! ✅

PR #68, opened 2024-04-21
Exported edit-mode utils

ewandank:
This little PR exports the util functions that were previously available under `@nebula.gl/edit-modes`. 
These are useful for implementing custom edit modes or extending upon existing ones. 

Please let me know if there is a different naming convention that would be preferred.

### DIFF
--- a/modules/editable-layers/src/index.ts
+++ b/modules/editable-layers/src/index.ts
@@ -36,6 +36,8 @@ import * as utils from './utils';
 
 export {utils};
 
+export {getPickedEditHandle, getEditHandlesForGeometry} from './edit-modes/utils';
+
 export type {EditMode} from './edit-modes/edit-mode';
 export type {GeoJsonEditModeType} from './edit-modes/geojson-edit-mode';
 export type {GeoJsonEditModeConstructor} from './edit-modes/geojson-edit-mode';


ibgreen (commented):
Thanks for the PR. Initial feedback with the hope that this can move forward quickly

- As you can see, we generally avoid * exports (for a variety of reasons) and explicitly list out the things we export, this PR should do the same.
- There are a lot of big functions in that file that we don't necessarily want to make part of the public interface and have to maintain. 
- Is there a specific function or two that are most useful?

ewandank:
Thanks for your quick review. 
I have updated this to export the 2 functions I am using, `getPickedEditHandle` and `getHandlesForGeometry`. 
If preferable, I can import them into the root `utils.ts` and then export them through that.

ibgreen (approved):
Approved. 

That said, normally there is a bit more to exposing new public APIs than just exporting them. If you were  a maintainer I would ask that:

- JSDoc/TSDoc style documentation on these two functions (so that we can see the docs in e.g. vscode when we "hover" over the functions in our code, and so that future maintainers get a clearly articulated understanding of what behavior they are expected to keep).
- A minimal test case? Even just importing the functions and making sure they is defined?
- At least an absolute minimum of documentation mentioning them?
-  Add a note to the module's what's new page mentioning them?

PR #72, opened 2024-05-01
Fix examples/arrow-layers volta path

birkskyum:
The examples in the examples/arrow-layers need an extra `../` in the volta path to reach the root


ibgreen (approved):
I wasn't aware there was a volta spec in each directory. I doubt that it is needed, but don't see the harm.

PR #73, opened 2024-05-01
examples/editable-layers/advanced: use vite/ts 5

birkskyum:
These examples can be run with latest vite/typescript


ibgreen (approved):
Should this be changed in all examples?

PR #75, opened 2024-05-01
Use maplibre in examples/editable-layers

birkskyum:
This allow running all the /examples/editable-layers/ without api keys.


ibgreen (approved):
Nice, thanks!

Issue #38, opened 2024-04-03
Tracker: editable-layers module


felixpalmer:
## Wishlist

- Goal: allow editable layers to be used with non-geospatial coordinates
- Goal: make this a generic layer pack (non-geo-speficic).
- Goal: Expose edit modes through Widgets somehow.

## v9.3 

- **High level**
  - [x] Docs - make sure the docs get the reader up to speed on the right things.
  - [x] `MeshLayer` - is it still needed? Docs says it should be [replaced](https://visgl.github.io/deck.gl-community/docs/modules/editable-layers/api-reference/layers/mesh-layer#lightsettings-object-optional)
  - [x] `Toolbox` React component - Could it be an `EditorLayer` composite layer with a suite of widgets?
- **examples** - Make all nebula.gl examples build again, or remove them
  - [x] editable-layers/sf (⚠️ renders, but throws errors)
  - [x] editable-layers/no-map
  - [x] editable-layers/getting-started
  - [x] editable-layers/codesandbox
  - [x] React 18, functional components, etc
- **Website** - 
  - [ ] editable-layers/editable-h3-cluster-layer
  - [x] editable-layers/sf (⚠️ renders, but throws errors)
  - [ ] editable-layers/overlays
  - [x] editable-layers/no-map
  - [x] editable-layers/getting-started
  - [x] editable-layers/codesandbox
 

## v9.0 - Initial forking from nebula.gl

 - [ ] **Website** - 
  - [x] Add `editor` example #76 @birkskyum 
  - [x] Add `advanced` example 
  - [ ] editable-layers/editable-h3-cluster-layer
  - [ ] editable-layers/sf (⚠️ renders, but throws errors)
  - [ ] editable-layers/overlays
  - [x] editable-layers/no-map
  - [x] editable-layers/getting-started
  - [x] editable-layers/codesandbox
- [ ] *examples** - Make all nebula.gl examples build again
  - [x] editable-layers/editor 
  - [x] editable-layers/advanced
  - [x] editable-layers/editable-h3-cluster-layer
  - [x] editable-layers/sf (⚠️ renders, but throws errors)
  - [ ] editable-layers/overlays
  - [x] editable-layers/no-map
  - [x] editable-layers/getting-started
  - [x] editable-layers/codesandbox
  - [x] Maplibre, token free basemaps, etc @birkskyum 
  - [x] React 18, functional components, etc
- [x] **Remove `@deck.gl-community/react-editable-layers`** #96 
[...]

PR #76, opened 2024-05-01
 Add nebula examples to website

birkskyum:
Related to:
- #38 

Screenshot 2024-05-02 at 01 28 47

I also removed the unused images from the deck.gl examples

Note: In "Advanced" the "Draw 90 degrees polygon" prompts an error about missing coord, so it's commented out for now and should be investigated further.


website/docusaurus.config.js
```diff
@@ -50,28 +50,37 @@ const config = {
...
+            '@deck.gl-community/graph-layers': resolve('../modules/graph-layers/src'),
+            '@deck.gl-community/react': resolve('../modules/react/src'),
+            '@deck.gl-community/layers': resolve('../modules/layers/src'),
+            '@deck.gl-community/arrow-layers': resolve('../modules/arrow-layers/src'),
+            '@deck.gl-community/editable-layers': resolve('../modules/editable-layers/src'),
+            '@deck.gl-community/react-editable-layers': resolve('../modules/react-editable-layers/src'),
+            'react-deck-graph-layers': resolve('../modules/react-graph-layers/src'),
             react: resolve('node_modules/react'),
             'react-dom': resolve('node_modules/react-dom'),
+            '@deck.gl/aggregation-layers': resolve('../node_modules/@deck.gl/aggregation-layers'),
```
ibgreen, line 64: Should we make the community website buildable standalone? If we pull in checked out deck.gl during builds we may not get predictable publishing. 

We could comment these out so that they can be easily activated during development

website/package.json
```diff
@@ -31,12 +31,13 @@
     "d3-scale": "^3.2.1",
     "expr-eval": "^2.0.2",
     "mapbox-gl": "^2.0.0",
```
ibgreen, line 33: How hard would it be to drop mapbox completely here? Are there a few more examples in the repo that need to be updated?

website/src/examples/editable-layers-advanced.js
```diff
@@ -0,0 +1,30 @@
+import React, {Component} from 'react';
```
ibgreen, line 1: - Given that we will have a number of submodules, should we adopt a directory structure, `examples/editable-layers/advanced.js`?
- Can we use typescript instead of .js?

ibgreen (approved):
Spectacular, thanks for doing this!

website/package.json
birkskyum: Not hard, I'll look into it.

website/src/examples/editable-layers-advanced.js
birkskyum: It's a good idea. I actually started doing that and realized it required a bit more refactor of the example list component than I hoped for to make it nested. I can look into it either here or in a follow up.

ibgreen:
@birkskyum The new examples break the website build. Sometimes `yarn start` works but `yarn build` fails, so it is important to test an actual build.

```
cd website
yarn
yarn build
```

birkskyum:
> @birkskyum The new examples break the website build. Sometimes `yarn start` works but `yarn build` fails, so it is important to test an actual build.
> 
> 
> 
> ```
> 
> cd website
> 
> yarn
> 
> yarn build
> 
> ```

Oh okay, I'll make sure to pay attention to that

birkskyum:
Where are the ci liga?

ibgreen:
website is built and deployed manually at this time.

website/package.json
birkskyum: Fixed by
- #86

website/docusaurus.config.js
birkskyum: I'm still a quite confused about the purpose of this resolve() setup, and ocular in general, so I'd prefer if you made a pr with this relevant change.

website/src/examples/editable-layers-advanced.js
birkskyum: TS part resolved in
- #88 

Directory structure resolved by
- #89

PR #87, opened 2024-05-02
Remove unused files copied from deck.gl

birkskyum:


ibgreen (approved):
Nit: It would kind of be nice to still have a playground but set up with layers from this repo. But I we can always add it back later.

PR #88, opened 2024-05-02
Convert examples to typescript

birkskyum:
Related to https://github.com/visgl/deck.gl-community/pull/76#discussion_r1587471639


website/docusaurus.config.js
```diff
@@ -50,7 +50,6 @@ const config = {
         resolve: {
           modules: [resolve('node_modules'), resolve('../node_modules')],
           alias: {
-            'examples': resolve('../examples'),
```
ibgreen, line 53: Why remove? Too confusing? Agree that less "magic" is sometimes better.

ibgreen (approved):

website/docusaurus.config.js
birkskyum: It was because typescript didn't pick up on the ocular alias, and debugging that complexity seemed like a very large task compared to adding a relative path..

PR #89, opened 2024-05-02
Organize examples in folders

birkskyum:
Related to https://github.com/visgl/deck.gl-community/pull/76#discussion_r1587471639


ibgreen (approved):
Thanks. 

PS - I did send you an invite to the repo in case you want to push branches directly instead of working through a fork.

PR #92, opened 2024-05-02
Fix editable-layers/editor example

birkskyum:
importing the react-editable-layers outside of the BrowserOnly caused the issue.


website/src/examples/editable-layers/editor.tsx
```diff
@@ -20,7 +19,10 @@ class EditorDemo extends Component {
 
     return (
       <BrowserOnly>
-        {() => <App {...otherProps} />}
+      {()=>{
+        const App = require('../../../../examples/editable-layers/editor/example');
```
ibgreen, line 23: I guess this works. I'd feel better if we pinned down and removed the dependency, but I guess we have bigger fish to fry. 

A comment explaining why we are doing such an unusual import would have been nice.

ibgreen (approved):

website/src/examples/editable-layers/editor.tsx
birkskyum: Resolved by
- #94

PR #95, opened 2024-05-03
Fix crash in 90 degree editable-layers example

birkskyum:
This example doesn't crash anymore, but it still doesn't add the polygon on completion.

### DIFF
--- a/examples/editable-layers/advanced/src/example.tsx
+++ b/examples/editable-layers/advanced/src/example.tsx
@@ -104,7 +104,7 @@ const ALL_MODES: any = [
       {label: 'Draw Point', mode: DrawPointMode},
       {label: 'Draw LineString', mode: DrawLineStringMode},
       {label: 'Draw Polygon', mode: DrawPolygonMode},
-      // {label: 'Draw 90° Polygon', mode: Draw90DegreePolygonMode},
+      {label: 'Draw 90° Polygon', mode: Draw90DegreePolygonMode},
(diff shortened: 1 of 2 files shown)


examples/editable-layers/advanced/src/example.tsx
```diff
@@ -870,7 +870,7 @@ export default class Example extends React.Component<
   onEdit = ({updatedData, editType, editContext}) => {
     let updatedSelectedFeatureIndexes = this.state.selectedFeatureIndexes;
 
-    if (!['movePosition', 'extruding', 'rotating', 'translating', 'scaling'].includes(editType)) {
+    if (!['movePosition', 'extruding', 'rotating', 'translating', 'scaling', 'updateTentativeFeature'].includes(editType)) {
```
birkskyum: this was very noisy

modules/editable-layers/src/edit-modes/draw-90degree-polygon-mode.ts
```diff
@@ -25,7 +25,7 @@ export class Draw90DegreePolygonMode extends GeoJsonEditMode {
     const {mapCoords} = props.lastPointerMoveEvent;
 
     let p3;
-    if (clickSequence.length === 1) {
+    if (clickSequence.length <= 1) {
```
birkskyum, line 28: first two clicks make a line, then the bearing etc. can be calculated form that.

ibgreen:
yarn lint fix

ibgreen (approved):

PR #100, opened 2024-05-03
chore(graph-layers): remove test-utils module, reorganize code

ibgreen:
- No need for a separate test-utils module
- Rename packages to align with repo conventions. 
- Update the example to deck.gl v9 (unfortunately does not yet run).
- Add license headers


donmccurdy:
This all looks great!! 👏🏻

ibgreen:
Yep, only problem is example is not yet working. We really need working examples on the website for each module, otherwise that module isn't ready to be included.

PR #110, opened 2024-07-29
feat(editable-layers): use the geojson types instead of custom types

giangbv1984:
- Use the types of geojson instead of custom types
- Fixed bug: missed the boxicons
image


examples/editable-layers/editor/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "@deck.gl-community/editor",
```
ibgreen: While it wouldn't really matter here, lets only use the npm organization scope for actual modules
```suggestion
  "name": "editable-layers/editor",
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -4,7 +4,17 @@
 
 // Geometry types
 
-export type Position = [number, number] | [number, number, number];
+import {
```
ibgreen: Any reason why we need to alias names? Can't we just re-export "as is" from the geojson module?

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -4,7 +4,17 @@
...
+  Point as PointGeo,
+  LineString as LineStringGeo,
+  Polygon as PolygonGeo,
+  MultiPoint as MultiPointGeo,
+  MultiLineString as MultiLineStringGeo,
+  MultiPolygon as MultiPolygonGeo,
+  Position as PositionGeo,
+} from 'geojson';
```
ibgreen:
```suggestion
export {
  Point,
  LineString,
  Polygon,
  MultiPoint,
  MultiLineString,
  MultiPolygon,
  Position
} from 'geojson';
```

ibgreen (commented):
I am temporarily confused. Looking at npm, the geojson package doesn't seem to be exporting any types?
- https://www.npmjs.com/package/geojson
- https://github.com/caseycesari/GeoJSON.js/blob/master/geojson.js

modules/editable-layers/src/utils/geojson-types.ts
giangbv1984: I have resolved it.

examples/editable-layers/editor/package.json
giangbv1984: Reverted file

giangbv1984:
Hi @ibgreen. Can you help me to review this PR? Thanks!

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -4,7 +4,26 @@
...
+
+export {
+  Point,
+  LineString,
+  Polygon,
+  MultiPoint,
+  MultiLineString,
+  MultiPolygon,
+  Position
+};
```
ibgreen:
```suggestion
} from 'geojson';
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -20,36 +39,6 @@ export type AnyCoordinates =
...
-  type: 'MultiLineString';
-  coordinates: MultiLineStringCoordinates;
-};
-
-export type MultiPolygon = {
-  type: 'MultiPolygon';
-  coordinates: MultiPolygonCoordinates;
-};
-
 export type Geometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
```
ibgreen, line 34: Unrelated to this PR
- Add a `/** */` TSDoc style comment for each new definition. 
- I am not sure I like the `Geometry` and `Polygonal` types as these are not following normal conventions. But changing them probably requires updating other code, so would be for a different PR.
- For one thing, `Geometry` is typically called `Feature`

modules/react/package.json
```diff
@@ -40,7 +40,8 @@
     "prop-types": "^15.8.1",
     "styled-components": "^4.4.1",
     "styled-react-modal": "^3.1.1",
-    "supercluster": "^8.0.1"
+    "supercluster": "^8.0.1",
+    "boxicons": "^2.1.4"
```
ibgreen, line 44: Why are we adding `boxicons`? Was it missing? Kind of unrelated to this PR...

ibgreen (approved):

modules/react/package.json
giangbv1984: Because, the modules/react/src/components/icon.tsx use `boxicons `. If we don't install this library, we have an error (You can see the description) when running test

modules/editable-layers/src/utils/geojson-types.ts
giangbv1984: This is your code. I only import the types of geojson.

ibgreen:
@giangbv1984 this is approved. If you can make it pass the linter (run `yarn lint fix` and push again) I will land this.

giangbv1984:
> @giangbv1984 this is approved. If you can make it pass the linter (run `yarn lint fix` and push again) I will land this.

I have runned `yarn lint fix` and fixed the import @types/geojson. But, we still have a lot of warning:
image

ibgreen:
Had to land this manually to fix the linter issue.

PR #115, opened 2024-08-02
fix(editable-layers): Make the implicit dependency to layers module explicit. #112

### LINKED ISSUE #112: Module editable-layers has implicit dependency to layers
The module editable-layers has no explicit dependency to module layers: https://github.com/visgl/deck.gl-community/blob/master/modules/editable-layers/package.json

In fact the segments-layer is importing the layers module: https://github.com/visgl/deck.gl-community/blob/646f8328d27d67d596f05c9154072051ec5cf4f0/modules/editable-layers/src/lib/layers/segments-layer.ts#L9

This causes a build failure for any project, that only imports `@deck.gl-community/editable-layers`, but not `@deck.gl-community/layers`. The workaround is to also import `@deck.gl-community/layers`, but this implicit dependency should be made explicit by defining layers as peerDepdendency for editable-layers.

cg-sat:
closes #112

### DIFF
--- a/modules/editable-layers/package.json
+++ b/modules/editable-layers/package.json
@@ -70,6 +70,7 @@
     "viewport-mercator-project": ">=6.2.3"
   },
   "peerDependencies": {
+    "@deck.gl-community/layers": "^9.0.2",
     "@deck.gl/core": "^9.0.12",
     "@deck.gl/extensions": "^9.0.12",
     "@deck.gl/geo-layers": "^9.0.12",
--- a/modules/editable-layers/package.json
+++ b/modules/editable-layers/package.json
@@ -70,7 +70,7 @@
     "viewport-mercator-project": ">=6.2.3"
   },
   "peerDependencies": {
-    "@deck.gl-community/layers": "^9.0.2",
+    "@deck.gl-community/layers": "^9.0.0",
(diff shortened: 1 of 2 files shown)


cg-sat:
@ibgreen Here's the first PR.

modules/editable-layers/package.json
```diff
@@ -70,6 +70,7 @@
     "viewport-mercator-project": ">=6.2.3"
   },
   "peerDependencies": {
+    "@deck.gl-community/layers": "^9.0.2",
```
felixpalmer: In deck.gl we have the peerDependencies defined more loosely, e.g. `"^9.0.0"`, any reason not to do that in this repo also?

cg-sat: @felixpalmer Thanks for the help. I relaxed the version constraint according to your suggestion.-

ibgreen (approved):

cg-sat:
@ibgreen I figured out, what was causing yarn to use version 1 behavior. By running `yarn set version stable` it writes a `packageManager` key into the root package.json. If this key isn't present, yarn seems to fall back to version 1 behavior. I added the change in the root package.json in my other MR #112 With that change I only get the necessary changes in the lock file.

Issue #114, opened 2024-08-02
module react has a dependency for react 16 or 17, while another dependency requires 18


cg-sat:
The react module requires `react` and `react-dom` to have major version 16 or 17:
https://github.com/visgl/deck.gl-community/blob/646f8328d27d67d596f05c9154072051ec5cf4f0/modules/react/package.json#L46

The dependency `styled-react-modal: ^3.1.1` itself requires react and react-dom to have major version 18:
https://github.com/AlexanderRichey/styled-react-modal/blob/60f42fc5bce647387e8e7e8c9e34881f7a63ee0b/package.json#L35

This leads to a dependency conflict in `npm install`.

I see the following two solutions for this:
1. Remove the dependencies to react and react-dom, because it comes as transient dependency.
2. Update the dependencies to react and react-dom to `^18`. This has the disadvantage, that any future version upgrades for `styled-react-modal` needs to be reflected in the minimum react version as well.

ibgreen:
@cg-sat Thanks for reporting. 

> Update the dependencies to react and react-dom to ^18. 

Yes. Since this repo is a set of modules building on top of deck.gl, and latest deck.gl/react is using `"react": "^18.2.0"`, I see no reason to deviate from that.

Now, as they name says, this repo is called `deck.gl-community` because it is maintained by the community, meaning that we mostly rely on contributors to make PRs. 

I am however happy to help review, land, and publish.

ibgreen:
Landed, pending #126

PR #116, opened 2024-08-02
fix(react): Remove transient dependencies from direct dependencies.

cg-sat:
Proposal 1 for #114 
@ibgreen I'm not familiar with the projects' policy how to maintain transient dependencies. If proposal 2 is desired, I can change the PR accordingly.


cg-sat:
I just saw, that the code actually imports stuff from `styled-components` and `react`. This means, that these two packages are not only transient dependencies but direct dependencies. This means they should also be made explicit.

modules/react/package.json
```diff
@@ -38,13 +38,12 @@
...
     "prop-types": "^15.8.1",
-    "styled-components": "^4.4.1",
+    "styled-components": "^6.1.1",
     "styled-react-modal": "^3.1.1",
     "supercluster": "^8.0.1"
   },
   "peerDependencies": {
-    "react": "^16.14 || ^17",
-    "react-dom": "^16.14 || ^17"
+    "react": "^18"
```
ibgreen: Let's follow deck.gl and use "^18.2.0": https://github.com/visgl/deck.gl/blob/master/modules/react/package.json#L42.

ibgreen (commented):
@cg-sat Thanks for the PR. Can you also do a search for "react" dependencies across the repo and update across the board. In particular, can you make a pass on the examples directory?

ibgreen:
As you can see in the log, you need to also commit the updated yarn.lock file when changing dependencies. Otherwise CI will reject the PR.

ibgreen:
@cg-sat, looks like we have a merge conflict.

modules/react/package.json
cg-sat: I changed it to `^18.2.0`. Also updated to examples, that were still requesting react 16.

ibgreen (approved):
> I changed it to ^18.2.0. Also updated to examples, that were still requesting react 16.

Perfect, will merge as soon as clean

cg-sat:
> @cg-sat, looks like we have a merge conflict.

Fixing it just now.

PR #136, opened 2024-09-04
Feat: delete mode for geojson editable layers

moharamfatema:
#### Issue
There is currently no way to delete features from a `FeatureCollection` via user interaction.

#### Solution
Introduce a `DeleteMode` class that extends `GeoJsonEditMode`. This mode will:
1. Handle user clicks to identify and remove the selected feature(s) from the `FeatureCollection`.
2. Trigger an `onEdit` callback to update the collection in the application state.

#### Implementation
- **delete-mode.ts**: Implement the `DeleteMode` class.
- **delete-mode.test.ts**: Add unit tests to ensure correct functionality.

This will enable users to manage their data more effectively by removing unwanted features.

### DIFF
--- a/modules/editable-layers/src/edit-modes/delete-mode.ts
+++ b/modules/editable-layers/src/edit-modes/delete-mode.ts
@@ -2,16 +2,17 @@ import {FeatureCollection} from '../utils/geojson-types';
 
 import {GeoJsonEditMode} from './geojson-edit-mode';
 import {ClickEvent, ModeProps} from './types';
-import {ImmutableFeatureCollection} from './immutable-feature-collection';
-
 export class DeleteMode extends GeoJsonEditMode {
   handleClick(_event: ClickEvent, props: ModeProps<FeatureCollection>): void {
     const selectedFeatureIndexes = props.lastPointerMoveEvent.picks.map((pick) => pick.index);
     if (selectedFeatureIndexes.length > 0) {
-      const updatedData = new ImmutableFeatureCollection(props.data)
-        // in case of overlapping features, delete only the most recent feature
-        .deleteFeatures([selectedFeatureIndexes[0]])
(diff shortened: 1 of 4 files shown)


modules/editable-layers/src/edit-modes/delete-mode.ts
```diff
@@ -0,0 +1,27 @@
...
+import {ClickEvent, ModeProps} from './types';
+import {ImmutableFeatureCollection} from './immutable-feature-collection';
+
+export class DeleteMode extends GeoJsonEditMode {
+  handleClick(_event: ClickEvent, props: ModeProps<FeatureCollection>): void {
+    const selectedFeatureIndexes = props.lastPointerMoveEvent.picks.map((pick) => pick.index);
+    if (selectedFeatureIndexes.length > 0) {
+      const updatedData = new ImmutableFeatureCollection(props.data)
+        // in case of overlapping features, delete only the most recent feature
+        .deleteFeatures([selectedFeatureIndexes[0]])
```
ibgreen: Nit: Doesn't it look weird to create an ImmutableFeatureCollection and then immediately proceed to mutate it :) Does this copy the entire collection? I am not up to speed on this code.

ibgreen (commented):
@moharamfatema Thanks for this contribution, I didn't even realize that we were missing this feature. 

Please make a pass on the docs as well.

ibgreen:
https://visgl.github.io/deck.gl-community/docs/CONTRIBUTING#fixing-ci-issues

modules/editable-layers/src/edit-modes/delete-mode.ts
moharamfatema: Good point, I reimplemented this by calling `filter` on the received data instead.

PR #138, opened 2024-09-14
Add Leaflet module

zakjan:


examples/leaflet/app.ts
```diff
@@ -0,0 +1,58 @@
+import {DeckLayer} from '@deck.gl-community/leaflet';
+import {MapView} from '@deck.gl/core';
+import {GeoJsonLayer, ArcLayer} from '@deck.gl/layers';
+import * as L from 'leaflet';
+import 'leaflet/dist/leaflet.css';
+
+// source: Natural Earth http://www.naturalearthdata.com/ via geojson.xyz
+const AIR_PORTS =
```
ibgreen, line 8:
```suggestion
const AIRPORTS_URL =
```

modules/leaflet/src/deck-layer.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  pickObjects(opts: Parameters<Deck['pickObjects']>[0]): ReturnType<Deck['pickObjects']> {
+    if (!this._deck) {
+      return [];
+    }
+
+    return this._deck.pickObjects(opts);
+  }
+
+  #getMap(): L.Map & {_animatingZoom: boolean; _getMapPanePos: () => L.Point} {
```
ibgreen, line 102: Nit: are these private declarations? We usually use the typescript private keyword and a leading underscore

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "type": "module",
   "workspaces": [
+    "examples/*",
```
ibgreen, line 15: Might be good to do this and the bing maps fix in separate PRs. This one in particular has caused some issues in the past. Make sure you can build the website. May have to revert if it messes up publishing.

modules/leaflet/test/index.spec.ts
```diff
@@ -0,0 +1,12 @@
+// deck.gl-community
```
ibgreen, line 1: Fine, though not super excited about files called `index...` maybe we could name it after the tests inside?

ibgreen (approved):
Very nice indeed. Some minor nits, feel free to ignore based on what makes sense to you. Plus a suggestion to break out some changes into separate PR.

ibgreen:
@zakjan Feel free to merge whenever you are ready, or let me know if you need me to merge.

PR #143, opened 2024-09-24
Enable blank issues

kylebarron:
I was planning to create a new issue to discuss whether arrow layers should accept Table or RecordBatch input. But currently the only way to create an issue is to fill out the full bug template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: I have a question / I need help
     url: https://github.com/visgl/deck.gl-community/discussions


ibgreen:
This is intentional. For repo hygiene reasons we refer non-confirmed bugs to the discussions tab.

kylebarron:
Can we mirror upstream deck.gl and have a feature request template? https://github.com/visgl/deck.gl/blob/master/.github/ISSUE_TEMPLATE/feature-request.yml

ibgreen:
Well, the thinking was that this repo contains a bunch of modules without active maintainers - so having a "submit feature request" like in deck seems disingenuous, it sends the wrong message, and we'll end up with a ton of feature requests in the issues tab that most likely won't get addressed unless some random contributor picks them up.

If opening a feature request in discussions doesn't work perhaps we need to reconsider.

kylebarron:
I like discussions for Q&A, but I don't like discussions for bugs or feature requests because they don't show interlinking as easily. Whenever you reference an issue or PR from any other issue on GitHub, you see from the target that it was linked to by the source PR. Discussions don't do that, and so it's harder to follow the graph of interconnected work.

ibgreen:
Replaced by #144

ibgreen:
Feature proposals have been added.

PR #147, opened 2024-10-16
Update peerdependencies in  package.json

kikonejacob:
The @deck.gl-community/layers is causing some dependencies problem when using a different version of deck gl.
This PR fixes the problem by using peerdependencies

### DIFF
--- a/modules/layers/package.json
+++ b/modules/layers/package.json
@@ -29,8 +29,8 @@
     "test": "vitest run",
     "test-watch": "vitest"
   },
-  "dependencies": {
-    "@deck.gl/core": "^9.0.12",
+  "peerDependencies": {
+   "@deck.gl/core": "^9.0.12",
     "@deck.gl/extensions": "^9.0.12",
     "@deck.gl/geo-layers": "^9.0.12",
     "@deck.gl/layers": "^9.0.12",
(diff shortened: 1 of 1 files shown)


ibgreen:
> The @deck.gl-community/layers is causing some dependencies problem when using a different version of deck gl. This PR fixes the problem by using peerdependencies

I suppose the downside of this is that the user becomes responsible for adding all those dependencies to their own package.json?

CC @Pessimistress who has carefully calibrated these things in our core repos.

kikonejacob:
> > The @deck.gl-community/layers is causing some dependencies problem when using a different version of deck gl. This PR fixes the problem by using peerdependencies
> 
> I suppose the downside of this is that the user becomes responsible for adding all those dependencies to their own package.json?
> 
> CC @Pessimistress who has carefully calibrated these things in our core repos.

yes. I think deckgl dependencies need to added dependency since there can only be one deck gl version in an app.
I don't know much about the luma dependencies but we can make them a  hard dependency if it is ok to have multiple version of the same package.

ibgreen:
>  if it is ok to have multiple version of the same package.

No it is not OK to have multiple version. But adding a long list of luma deps will make these packages quite inconvenient to use as the user has to add a dozen or more dependencies manually to their apps.

And, assuming all packages use flexible version specifiers, yarn or npm should be able to deduplicate them. Have you root caused what goes wrong in your install? Are you trying to use this with a specific older version of deck?

Issue #151, opened 2024-11-04
[Feat] add geometry's description when creating circle or ellipses using editable-layers


hadbn:
### Target Use Case

When creating vectors using editable-layers, it might be usefull to be able to characterize the drawn geometry, or even to re-draw it. If it is easy to get back to the parameters of the geometry for squares or rectangle, for instance, it is much more expensive for circle and ellipses.
Thus, when a circle is created, its center and radius should be stored in the vector properties.
When an ellipse is created, we should store its center, its axis lengths, and the angle.

### Proposal

We should just propagate the properties along all the methods, as they are already calculated. I will propose a PR.

PR #154, opened 2024-11-14
Adds properties of the circles and ellipses created with editable layers

hadbn:
When drawing circles or ellipses properties of the created geometry are now stored in the vector's properties.

This feature is related with : https://github.com/visgl/deck.gl-community/issues/151


ibgreen (commented):
This seems reasonable. I feel that a feature like this needs to include 
- documentation updates
- at least some basic tests

hadbn:
@ibgreen I updated the documentation, added one test for DrawEllipseUsingThreePointsMode and updated the test of DrawCircleByDiameterMode.

Are you happy with that ? Do you have any other suggestion ?

hadbn:
Are you ok with those modifications ?

ibgreen (approved):
@hadbn Thanks for your patience and persistence! The code changes looks great. 

 My remaining hesitation is related to the storage of geometry specific fields in the properties. This does not seem well documented today. It makes it hard for me to review since I don't know how the code is supposed to work.

I.e. is this "polluting" the GeoJSON feature level property object (where applications normally store all the non-geometry aspects of the table row). 

```js
{
   type: 'Polygon',
   geometry: {...},
   properties: {population: 100, averageIncome: ..., ..., xSemiAxis: ...}
}
```

I wish there was a developer's guide page that described this (how GeoJSON objects are manipulated by the GeoJSON layer) so that it was clear how this is supposed to work.

ibgreen:
If we are adding a bunch of new properties to the feature-level properties object, maybe we should try to keep them in a single sub-object with a well-known name, so the that app can get rid of them in a predictable way once editing is done?

hadbn:
Sorry about that, I just pushed.

Concerning your question, from what I read, properties should not be added inside the `geometry field` and thus this should not be done :
`geometry: {type: 'Polygon', coordinates: [[[...]]], properties: {, xSemiAxis: ...}}`

As the field `shape` was already in the `properties` object, I chose to add other fields at the same level. However it could be possible to chose :

`{
   type: 'Feature',
   geometry: {type: 'Polygon', coordinates: [[[...]]]},
   properties: {shape: Ellipse, population: 100, averageIncome: ..., ..., geomProperties: {xSemiAxis: ..., ...}}
}`

It might even be possible to include `shape` inside `geomProperties` but it will impact the existing tests on the shape field along the code.

What do you prefer ?

ibgreen:
Thanks for indulging my "bike shedding".

> properties should not be added inside the geometry field and thus this should not be done :

Correct, it would not be standard GeoJSON. Agree that is a good reason to avoid it.

> It might even be possible to include shape inside geomProperties but it will impact the existing tests on the shape field that are made along the code.

- I think it would be reasonable to collect the editable-layers generated fields inside an object. I would probably lean towards naming it `editProperties` to highlight that it is related to editable-layers module.
- I would also store the `shape` property inside that object. 
- And we could still store it another copy of `shape` on the top-level object for backwards compatibility, but deprecate that usage in the upgrade guide.

hadbn:
@ibgreen The changes we discussed are done.
However, I'm not familiar with the upgrade guide. What should I do to let users know that `properties.shape` is now deprecated and replaced by `properties.editProperties.shape` ?

PR #179, opened 2024-12-12
Leaflet fixes

zakjan:


modules/leaflet/src/deck-utils.ts
```diff
@@ -4,7 +4,9 @@
 
 import * as L from 'leaflet';
 import {Deck} from '@deck.gl/core';
-import type {DeckProps, ViewStateMap, MapView} from '@deck.gl/core';
+import type {DeckProps, View, ViewStateMap, MapView} from '@deck.gl/core';
+
+export type ViewOrViews = View | View[] | null;
```
ibgreen, line 9: Nit: Quietly wondering why this is necessary to add here, and not the default in deck...  @Pessimistress

ibgreen (approved):

modules/leaflet/src/deck-utils.ts
zakjan: True, I had to add exactly because it's missing. It's not even a default value, default is null

Issue #150, opened 2024-10-29
Under certain conditions, ModifyMode can modify an incorrect vertex


rjwats:
### Module

- [ ] deck.gl-community/arrow-layers
- [ ] deck.gl-community/bing-maps
- [X] deck.gl-community/editable-layers
- [ ] deck.gl-community/experimental
- [ ] deck.gl-community/graph-layers
- [ ] deck.gl-community/layers
- [ ] deck.gl-community/react
- [ ] deck.gl-community/react-graph-layer

### Description

When modifying a geometry using ModifyMode there is an issue which can result in the wrong vertex being updated. This occurs when you "drop" a vertex close to another vertex (specifically one with a higher ordinal within coordinate list). 

I believe getPickedEditHandle in handleStopDragging is returning the vertex with the highest ordinal from the geometry, which is why the handleStopDragging modifieds the wrong vertex in this situation.

I'm not 100% confident in the correct implementation, however the following override appears to solve the issue for me:

```typescript
export default class CustomModifyMode extends ModifyMode {

  handleStopDragging(event: StopDraggingEvent, props: ModeProps<FeatureCollection>) {
    const selectedFeatureIndexes = props.selectedIndexes;
    const editHandle = getPickedEditHandle(event.pointerDownPicks);

    if (selectedFeatureIndexes.length && editHandle) {
      this._dragEditHandle('finishMovePosition', props, editHandle, event);
    }
  }

}
```

Attached is a video of the issue. See the below codesandbox for an interactive example of the problem with instructions and a visual aid to help observe the problem during editing:

https://codesandbox.io/p/sandbox/fwhvn7

https://github.com/user-attachments/assets/ecdc7d68-c341-4f6d-8807-9ffa064a9ce4

Thanks :)

### Expected Behavior

Only the dragged vertex should be updated to reflect the location where the user has dropped it.

### Steps to Reproduce

Open the codesandbox:

https://codesandbox.io/p/sandbox/fwhvn7

Drag the left most vertex and place it so it overlaps one of the other vertexes. Observe how the the vertex you have dropped the point onto is modified unexpectedly. I believe getPickedEditHandle in handleStopDragging is returning the vertex with the highest ordinal from the linestring, which is why the handleStopDragging modifieds the wrong vertex in this situation.

### Environment

- Framework version: @deck.gl-community/editable-layers 9.0.3
- Browser: Any
- OS: Any

timnyborg:
[...]

PR #198, opened 2025-01-10
Fix issue where modify mode can update the wrong vertex

rjwats:
Possible fix for https://github.com/visgl/deck.gl-community/issues/150 where ModifyMode can update the wrong vertex in the feature.

The regression test will fail without the fix in place, the test output demonstrating how the wrong vertex is be updated when using "picks" instead of "pointerDownPicks":

image (1)

### DIFF
--- a/modules/editable-layers/src/edit-modes/modify-mode.ts
+++ b/modules/editable-layers/src/edit-modes/modify-mode.ts
@@ -284,7 +284,7 @@ export class ModifyMode extends GeoJsonEditMode {
 
   handleStopDragging(event: StopDraggingEvent, props: ModeProps<FeatureCollection>) {
     const selectedFeatureIndexes = props.selectedIndexes;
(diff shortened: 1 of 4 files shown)


timnyborg:
Just to note: I'm certain this is the correct solution - it's identical to one I opened 2 years ago on nebula.gl (https://github.com/uber/nebula.gl/pull/868/files) and we've had patched in our applications ever since.

kevduong1 (approved):

charlieforward9 (changes requested):
Hey @rjwats, thanks for making this contribution. Would you be able to update this branch on master so we can move forward with merging it in, I am doing a package-wide issue/PR triage and whatever help I can get with syncing these, the best. If not, please let me know so I can prioritze this among the other tasks.

charlieforward9 (approved):
test

Issue #201, opened 2025-01-22
Drawing modes not working [Bug]


PacoVerduBravo:
### Module

- [ ] deck.gl-community/arrow-layers
- [ ] deck.gl-community/bing-maps
- [x] deck.gl-community/editable-layers
- [ ] deck.gl-community/experimental
- [ ] deck.gl-community/graph-layers
- [ ] deck.gl-community/layers
- [ ] deck.gl-community/react
- [ ] deck.gl-community/react-graph-layer

### Description

None of the drawing modes seem to be functioning as expected. When configuring the onEdit function, the editType parameter never arrives as addFeature. Additionally, in many cases, the onEdit function is not executed at all.

### Expected Behavior

All drawing modes should work as expected

### Steps to Reproduce

Unfortunately, I couldn't provide a CodeSandbox example because I encountered errors related to dependency setup
```
import React from 'react';
import DeckGl from '@deck.gl/react';
import Map from 'react-map-gl';
import {
  EditableGeoJsonLayer,
  ViewMode,
  DrawRectangleUsingThreePointsMode,
  DrawPointMode,
  DrawRectangleFromCenterMode,
  DrawPolygonMode,
  DrawLineStringMode,
  MeasureDistanceMode,
  Draw90DegreePolygonMode
} from '@deck.gl-community/editable-layers';

const MisionMap = () => {
  const MAPBOX_ACCESS_TOKEN =  " "
  const mapsStyles = {
    Dark: 'mapbox://styles/mapbox/dark-v10',
    Satelite: 'mapbox://styles/mapbox/satellite-v9',
    Positron: 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json'
  };
  const INITIAL_VIEW_STATE = {
    longitude: -3.3052,
    latitude: 42.4247,
    zoom: 13,
    maxPitch: 70,
    pitch: 5,
    bearing: 0
  };
  const [features, setFeatures] = React.useState({
    type: 'FeatureCollection',
    features: []
  });
  const [mode, setMode] = React.useState(() => DrawPointMode);
  const [selectedFeatureIndexes] = React.useState([]);

  const layer = new EditableGeoJsonLayer({
    data: features,
    mode,
    selectedFeatureIndexes,
    onEdit: ({ editType, updatedData }) => {
      console.log(editType);
      setFeatures(updatedData);
    }
  });
  return (
    <DeckGl
      initialViewState={INITIAL_VIEW_STATE}
      controller={{
[...]

PR #162, opened 2024-11-26
chore: Upgrade to deck.gl v9.1

ibgreen:
Some API changes in mjolnir affecting editable-layers perhaps.

image

PR #203, opened 2025-02-02
[docs] Example overlaid on v9.1

charlieforward9:
Uses changes from #162 to create an example of how to use `EditableGeoJsonLayer` using [the overlaid approach](https://deck.gl/docs/developer-guide/base-maps/using-with-maplibre#example-interleaved-or-overlaid), demonstrating a bug explained in https://github.com/visgl/deck.gl-community/issues/201, where the drawing modes do not work. 

Please review and advise.


charlieforward9:
After many hours attempting to learn and resolve the issue, I must step back and absorb the defeat of the day. 

I tried bumping mjolnir to v3 and messing with it in this method:
https://github.com/visgl/deck.gl-community/blob/8b5c3a9f8474f729195995e4dbe5556eb1ef0d1c/modules/editable-layers/src/editable-layers/editable-layer.ts#L80-L92

No matter my approach, the events would not fire... please enlighten me 😩

ondave:
Having exactly the same problem.

I think the root cause is that the anyclick event does not get recognised. This gets it working (for me anyway, no gauarntee it works in the generic case on all devices):
```

_addEventHandlers() { 
   // @ts-expect-error accessing protected props 
   const {eventManager} = this.context.deck; 
   const {eventHandler} = this.state._editableLayerState; 
  
   const _EVENT_TYPES=EVENT_TYPES.concat(['click'])
   for (const eventType of EVENT_TYPES) { 
     eventManager.on(eventType as any, eventHandler, { 
       // give nebula a higher priority so that it can stop propagation to deck.gl's map panning handlers 
       priority: 100 
     }); 
   } 
 }

_onclick(eventSrc){
   this._onanyclick(eventSrc);
}
```

charlieforward9:
@ondave this snippet is not working for me, can you contribute it to my PR so I can review it?

ondave:
Ah, yes, sorry, didn't read your exact use case close enough. I was having issues with a vanilla deck.gl setup, and that fix patches that case OK, where anyclick just doesn't get fired.

For the MapboxOverlay case, I don't think the click events ever get to the deck.gl eventManager. The MapboxOverlay captures the events and forwards the events directly to the deck.gl _onEvent function, but the EditableLayers events are registered in the deck.gl eventManager.

Just out of interest did the EditableLayers (formerly nebula.gl, and pre deck v9) ever work with the MapboxOverlay?

I think one of the core dev team would have a better idea about all this.

charlieforward9:
> For the MapboxOverlay case, I don't think the click events ever get to the deck.gl eventManager

It seems so, hence why your snippet did not work for me.

When I call onClick within the `<Map />` component, the map receives and fires the event, but yeah in the overlaid layers, nothing...

> Just out of interest did the EditableLayers (formerly nebula.gl, and pre deck v9) ever work with the MapboxOverlay?

I am not sure of this, but this is what I would like to get working in v9.1+.

laurence-hudson-mindfoundry:
@charlieforward9 Have you had any progress finding a solution to this? I appear to be getting bitten by the same issue (EditableGeoJsonLayer not working if following the overlay approach); as speculated above I dont think this related to the v9.1/mjolnir.js change.

laurence-hudson-mindfoundry:
> @charlieforward9 Have you had any progress finding a solution to this? I appear to be getting bitten by the same issue (EditableGeoJsonLayer not working if following the overlay approach); as speculated above I dont think this related to the v9.1/mjolnir.js change.

@charlieforward9 I've had a breakthrough for my own map. Using the interleaved overlay rendering variant (interleaved prop on MapboxOverlay) avoids the event swallowing. While interleaved has some consequences, at least for my app there bearable. So functioning EditableGeoJsonLayer while also retaining the react-map-gl as the outer component (for maplibre's NavigationControl / map api / etc).

charlieforward9:
Thanks for the update. Are you able to reproduce it in my example repo that this PR uses so I can build from that?

laurence-hudson-mindfoundry:
Here's a minimal example using MapOverlay (interleaved true works, false doesn't)

charlieforward9:
I will spend some time tomorrow looking into this just a little bit more, if I am unable to figure it out, I will attempt to move forward with `interleaved={true}` and refactor this PR to include an interleaved example and await community help on resolving the overlaid issue.

More to come tomorrow.

charlieforward9:
@ibgreen this example should be ready to get merged in

PR #204, opened 2025-02-08
Update to deck 9.1

ewandank:
- Updated the deck peer dependency to `9.1.0`
- Updated the `deck.gl-community/packages/*` version to `9.1.0-beta.0`. If preferred i can change this to `9.1.0`.
-  Updated the raster layer to use the `updateModuleSettingsWebGL` instead of the previous `updateModuleSettings`
- Fixed a type error in the rounded rectangle layer to pass through all draw options, instead of just the uniforms. 
-  Changed the editable layer to listen to the `click` event rather then the `anyclick` event as `mjolnir.js@3.0.0` appears to no longer emit the `anyclick` event.


felixpalmer:
@ewandank see related changes here: https://github.com/visgl/deck.gl-community/pull/205

PR #205, opened 2025-02-10
Updates editable-layers & graph-layers to UBOs

felixpalmer:
## Changes

- Bump to `9.1.0` following same update as used in https://github.com/visgl/deck.gl-community/pull/162
- Update `RoundedRectangleLayer` to use UBOs
- Update `EditablePathLayer` to use UBOs
- Update `color` `ShaderModule` to use UBOs
- For now apply [quick-fix](https://github.com/visgl/deck.gl-community/commit/c6014498dd22575668c85dd1d4391ef76629b90e) to raster layers to fix build

@ibgreen the build is working with these changes, but I'm not having any luck running the examples, could you let me know which examples are known to work so I can better test?


ibgreen:
> could you let me know which examples are known to work so I can better test?

Good question, keeping the examples alive over time has been a challenge, the few that are included by the website should have been working.
https://github.com/visgl/deck.gl-community/blob/master/website/src/examples-sidebar.js

PR #206, opened 2025-02-10
Update `editable-layer` to use `click` instead of `anyclick` event

ewandank:
- Updated the deck to  `9.1.0` in line with #205 
-  Changed the editable layer to listen to the `click` event rather then the `anyclick` event as `mjolnir.js@3.0.0` appears to no longer emit the `anyclick` event.


modules/editable-layers/src/editable-layers/editable-layer.ts
```diff
@@ -15,7 +15,7 @@ import {
 } from '../edit-modes/types';
 import {Position} from '../utils/geojson-types';
 
-const EVENT_TYPES = ['anyclick', 'pointermove', 'panstart', 'panmove', 'panend', 'keyup'];
```
felixpalmer, line 18: @Pessimistress is this expected?

Pessimistress: mjolnir v3 no longer registers a default set of gesture recognizers. The recognizers used by deck are defined here:

https://github.com/visgl/deck.gl/blob/623265907fe2db31a91c9d2965e73c73cbf7efc1/modules/core/src/lib/constants.ts#L96-L102

felixpalmer:
@ewandank could you resolve the conflicts? You don't need the version bumps in this PR now that https://github.com/visgl/deck.gl-community/pull/205 is merged

felixpalmer:
@ewandank how have you tested this? I'm unable to run any examples or the website :/

felixpalmer (approved):
Got it running locally, works well. Thank you!

PR #215, opened 2025-02-13
Update dev tools

Pessimistress:
- Use latest dev tools
- Single tsconfig.json
- CI-based release workflow


tsconfig.json
```diff
@@ -26,6 +33,27 @@
       "deck-graph-layers": ["modules/graph-layers/src"]
     },
     "typeRoots": ["./node_modules/@types"],
+    "plugins": [
+      {
```
ibgreen, line 37: Feels like this could be an "extend" of some sort, to avoid duplication?

ibgreen (approved):

tsconfig.json
Pessimistress: You complained about dev-tools being a black box, and I agreed. I'd rather see all the transforms performed on the code without digging into some nested dependency.

PR #219, opened 2025-02-27
Fix EditableGeoJsonLayerProps typing, expose SelectionLayerProps

timnyborg:
Pulls together a few fixes I had open nebula.gl, that were approved but never merged or released:

https://github.com/uber/nebula.gl/pull/873
https://github.com/uber/nebula.gl/pull/858
https://github.com/uber/nebula.gl/pull/777

### DIFF
--- a/modules/editable-layers/src/editable-layers/editable-geojson-layer.ts
+++ b/modules/editable-layers/src/editable-layers/editable-geojson-layer.ts
@@ -118,7 +118,7 @@ export type EditableGeoJsonLayerProps<DataT = any> = EditableLayerProps & {
   mode?: any;
   modeConfig?: any;
   selectedFeatureIndexes?: number[];
-  onEdit?: (updatedData?, editType?: string, featureIndexes?: number[], editContext?) => void;
+  onEdit?: (editAction: EditAction<DataT>) => void;
 
   pickable?: boolean;
   pickingRadius?: number;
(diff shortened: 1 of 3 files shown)


charlieforward9 (commented):
Hey @timnyborg, thanks for making this contribution. Would you be able to update this branch on master so we can move forward with merging it in, I am doing a package-wide issue/PR triage and whatever help I can get with syncing these, the best. If not, please let me know so I can prioritze this among the other tasks.

modules/editable-layers/src/editable-layers/editable-geojson-layer.ts
```diff
@@ -119,7 +119,7 @@ export type EditableGeoJsonLayerProps<DataT = any> = EditableLayerProps & {
   mode?: any;
   modeConfig?: any;
   selectedFeatureIndexes?: number[];
-  onEdit?: (updatedData?, editType?: string, featureIndexes?: number[], editContext?) => void;
+  onEdit?: (editAction: EditAction<DataT>) => void;
```
charlieforward9, line 123: this is a good catch, surprised it made it all the way into a release like this and wasn't mentioned sooner or caught by a test. Lets try to add a test in here to maintain integrity. 

Something as simple as https://github.com/visgl/deck.gl-community/blob/d0c84e82248cd4e213a15205e7e34bd57ae578e4/modules/editable-layers/test/lib/layers/elevated-edit-handle-layer.spec.ts could do the job.

timnyborg: I don't think it was catchable, since it's just a misleading but still functional type that I stumbled on at random (its first argument was an implicit `any`, which `EditAction` happily satisfied).  Testing for a 'too broad type' isn't really possible, but I've just added some more explicit typing where it's invoked (getProps), so that the relationship is enforced by the type system.

charlieforward9 (approved):

PR #225, opened 2025-03-27
[editable-layers] Real double-click used to finish drawing

mwilniewiec:
Drawing of Polygon and LineString can now be finished using a normal double click.
Before this change drawing was finished by "Enter" key or by "slow double click" which wasn't recognized as a normal "double click".
This was a regress to old nebula.gl and older deck which I was using earlier.

### DIFF
--- a/modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
+++ b/modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
@@ -10,7 +10,8 @@ import {
   ModeProps,
   GuideFeatureCollection,
   TentativeFeature,
-  GuideFeature
+  GuideFeature,
+  DoubleClickEvent
 } from './types';
 import {Polygon, FeatureCollection} from '../utils/geojson-types';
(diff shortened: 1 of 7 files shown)


colehart:
I'm having this issue as well, and hoping this fix can make it into the next minor / patch release.

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
```diff
@@ -83,6 +84,23 @@ export class DrawPolygonMode extends GeoJsonEditMode {
...
+  finishDrawing(props: ModeProps<FeatureCollection>) {
+    const clickSequence = this.getClickSequence();
+    if (clickSequence.length > 2) {
+      const polygonToAdd: Polygon = {
+        type: 'Polygon',
+        coordinates: [[...clickSequence, clickSequence[0]]]
+      };
+
+      this.resetClickSequence();
+
```
felixpalmer:
```suggestion
```

colehart:
what release is this fix being included in?

PR #255, opened 2025-06-11
fix: prevent removing corner of a rectangle when lockRectangles is used

AnJ-K:
Added a simple check, similar to the one just below for hiding intermediate handles.
https://github.com/visgl/deck.gl-community/blob/b67a0b1635fa2a9831706ac1d8884cf554890531/modules/editable-layers/src/edit-modes/modify-mode.ts#L172-L177

When `lockRectangles: true` it is no longer possible to click on a corner of a rectangle to remove it.


AnJ-K:
@charlieforward9 yes, I'm still using it. I hadn't have time to revisit this PR, but I will try updating it during upcoming week.

charlieforward9:
Thank you! Looks like it merges cleanly right now so I just need to be rebased and to make sure this fix has a test validating it's functionality.

Issue #259, opened 2025-07-04
[Bug] @deck.gl-community/editable-layers missing @turf/rhumb-* dependencies


dulldrums:
### Module

- [ ] deck.gl-community/arrow-layers
- [ ] deck.gl-community/bing-maps
- [x] deck.gl-community/editable-layers
- [ ] deck.gl-community/experimental
- [ ] deck.gl-community/graph-layers
- [ ] deck.gl-community/layers
- [ ] deck.gl-community/react
- [ ] deck.gl-community/react-graph-layer

### Description

I ran into this bug while trying to build a project that uses pnpm as a package manager, and it has hoisting disabled. 

See the attached error message.

After investigation, it appears that the file [translate-from-center.ts](https://github.com/visgl/deck.gl-community/blob/master/modules/editable-layers/src/utils/translate-from-center.ts#L6-L8) is importing `@turf/rhumb-bearing`, `@turf/rhumb-distance`, and `@turf/rhumb-destination`. These are all missing from `@deck.gl-community/editable-layers`'s [package.json](https://github.com/visgl/deck.gl-community/blob/b67a0b1635fa2a9831706ac1d8884cf554890531/modules/editable-layers/package.json#L60).

As I have an easy means to test this fix out, I'm happy to make the changes and PR.

### Expected Behavior

The project should build!

### Steps to Reproduce

I just managed to replicate this in a private repository, and due to it's nature it might be difficult to setup on codepen/replit/etc. I'm happy to create a replicate public repo if required.

- Create new project that uses pnpm
- Set up pnpm to disable hoisting
- Add @deck.gl-community/editable-layers@9.0.3 as a dependency
- `pnpm install`
- `tsc && vite build`

### Environment

- Framework version: @deck.gl-community/editable-layers@9.0.3
- Browser: n/a
- OS: osx 15.5

### Logs

```bash
error during build:
[vite]: Rollup failed to resolve import "@turf/rhumb-bearing" from "/Users/dulldrums/workspace/deck.gl-community-pnpm-test/node_modules/.pnpm/@deck.gl-community+editable-layers@9.0.3_@deck.gl-community+layers@9.0.3_@deck.gl+core@9.1.12_xelhxzf37u2tgjh3d3owfuqq54/node_modules/@deck.gl-community/editable-layers/dist/utils/translate-from-center.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
[...]

PR #260, opened 2025-07-04
bug(editable-layers): Add missing @turf-rhumb* dependencies

dulldrums:
This addresses https://github.com/visgl/deck.gl-community/issues/259

Adds missing "@turf/rhumb-bearing", "@turf/rhumb-distance", and "@turf/rhumb-destination" dependencies to editable-layers.


charlieforward9:
With https://github.com/visgl/deck.gl-community/pull/447 nearing merge. This PR will be deprecated due to the incorrect version. If this is still an issue, open a new PR with the correct version of turf-rhumb.

Issue #273, opened 2025-07-15
[Bug] peerDependencies for deck/luma/loaders should pin to minor version (rather than major)


keller-mark:
### Module

- [ ] deck.gl-community/arrow-layers
- [ ] deck.gl-community/bing-maps
- [x] deck.gl-community/editable-layers
- [ ] deck.gl-community/experimental
- [ ] deck.gl-community/graph-layers
- [x] deck.gl-community/layers
- [ ] deck.gl-community/react
- [ ] deck.gl-community/react-graph-layer

### Description

When trying to use `@deck.gl-community/editable-layers` with a particular minor version of deck.gl, my package manager always tries to override it to use a newer minor version, because the peer dependencies pin to the major versions (e.g., `"@deck.gl/core": "^9.1.12"` rather than the minor version which can be done with the tilde `"@deck.gl/core": "~9.1.12"`)

https://github.com/visgl/deck.gl-community/blob/e88211052d3ecba78a5982ee581d6c7b95008b65/modules/editable-layers/package.json#L75

This causes issues, as minor versions of deck.gl and luma.gl can introduce breaking changes to the internals needed for such custom layers.

### Expected Behavior

Pin to minor versions of deck/lumaGL packages using tilde version syntax.

### Steps to Reproduce

Try to install a past version such as `"@deck.gl-community/editable-layers": "9.0.3"` with its matching DeckGL version 

See https://unpkg.com/@deck.gl-community/editable-layers@9.0.3/package.json , this version specified `"@deck.gl/core": "^9.0.12"` in its package.json . But installing today, my package manager forces me to install @deck.gl/core v9.1.x which is incompatible.

### Environment

- Framework version: v9.x
- Browser: N/A
- OS: macOS

ibgreen:
For reference, based on description, it looks like this is a problem when using with deck.gl v9.0
- Minor version pinning should be in place for deck.gl-community v9.1 and v9.2 - if that isn't the case we will fix.
- We normally only provide support for one back release, and unfortunately, the 9.0-release branch likely does not have updated publishing tooling, so there may not be a quick fix.
Not sure if it helps, but you may be able to use the "resolutions" field in your package.json to constrain the module versions until you are able to upgrade to a recent deck.gl version.

PR #275, opened 2025-08-13
Pin deck.gl and luma.gl dependencies to their minor versions (rather than major)

keller-mark:
Fixes https://github.com/visgl/deck.gl-community/issues/273


modules/editable-layers/package.json
```diff
@@ -71,15 +71,15 @@
...
-    "@deck.gl-community/layers": "^9.1.0-beta.2",
-    "@deck.gl/core": "^9.1.12",
-    "@deck.gl/extensions": "^9.1.0",
-    "@deck.gl/geo-layers": "^9.1.0",
-    "@deck.gl/layers": "^9.1.0",
-    "@deck.gl/mesh-layers": "^9.1.0",
-    "@luma.gl/constants": "^9.1.0",
-    "@luma.gl/core": "^9.1.0",
-    "@luma.gl/engine": "^9.1.0",
+    "@deck.gl-community/layers": "~9.1.0-beta.2",
```
keller-mark, line 74: Is this correct? Is `@deck.gl-community/layers` versioned in the same way, where minor versions have breaking changes?

modules/raster-layers/package.json
```diff
@@ -32,11 +32,11 @@
     "lodash.isequal": "^4.5.0"
   },
   "peerDependencies": {
-    "@deck.gl/core": "^9.1.12",
-    "@deck.gl/layers": "^8.8.0",
-    "@deck.gl/mesh-layers": "^8.8.0",
-    "@luma.gl/core": "^8.2.0",
-    "@luma.gl/engine": "^8.2.0"
+    "@deck.gl/core": "~9.1.12",
+    "@deck.gl/layers": "~8.8.0",
```
keller-mark, line 36: Is this module intended to be using both v8 and v9?

keller-mark:
Seems to have been fixed by https://github.com/visgl/deck.gl-community/commit/5cb9bfa2a656881298ecb30ee621c796882b5354#diff-2386931ff72ff4a107645241eb8b19263121b33d8c86ca8640f5958b57bc986b

modules/raster-layers/package.json
chrisgervang: That seems more like an indicator that the module isn't being maintained by anyone. I wouldn't expect these major versions to be compatible

PR #284, opened 2025-10-15
chore: Bump to deck.gl@9.2

ibgreen:
We need a version of deck.gl-community that is compatible with deck.gl 9.2.
I just pushed a 9.1-release branch with current master.


modules/raster-layers/dist.index.cjs.map
chrisgervang, line 1: Why do we track a source map in git?

modules/graph-layers/src/widgets/view-control-widget.tsx
```diff
@@ -230,17 +230,17 @@ export class ViewControl extends Component<ViewControlProps> {
...
   viewId?: string | null = null;
   viewports: {[id: string]: Viewport} = {};
-  deck?: Deck<any>;
   element?: HTMLDivElement;
+  className = 'deck-widget-view-control';
 
   constructor(props: ViewControlProps) {
+    super(props);
     this.props = {...ViewControl.defaultProps, ...props};
     this.id = props.id || 'zoom';
```
chrisgervang, line 245: The abstract class sets id, style, and className now

modules/graph-layers/src/widgets/view-control-widget.tsx
```diff
@@ -284,6 +275,17 @@ export class ViewControlWidget implements Widget<ViewControlProps> {
...
+        zoomBy={this.handleDeltaZoom.bind(this)}
+        panBy={this.handlePanBy.bind(this)}
+      />
+    );
+    render(ui, rootElement);
+  }
+
   setProps(props: Partial<ViewControlProps>) {
     Object.assign(this.props, props);
   }
```
chrisgervang, line 291: Should call super, or be removed if we aren't customizing it

modules/graph-layers/src/widgets/view-control-widget.tsx
```diff
@@ -250,7 +250,6 @@ export class ViewControlWidget implements Widget<ViewControlProps> {
     // props.zoomInLabel = props.zoomInLabel || 'Zoom In';
     // props.zoomOutLabel = props.zoomOutLabel || 'Zoom Out';
     props.style = props.style || {};
-    this.props = props;
   }
 
   onAdd({deck}: {deck: Deck<any>}): HTMLDivElement {
```
chrisgervang, line 255: I think a good portion of this is handled by the abstract class now.

This function also adds a class "deck-widget-zoom" (which suggests it's a fork of ZoomWidget). That should be a good reference then for how this widget should be written in 9.2

chrisgervang:
Fixes #273?

PR #379, opened 2025-11-08
feat (timeline-layers): example

charlieforward9:
In need of a big-data timeline for a project that can support hundreds of thousands of datapoints. Current solution is DOM-driven and lags out with a few hundred. 

A timeline solution as a deck.gl composite layer introduces a performant interactive no-low-code tool supporting direct track integration with all Deck props including projection, viewState, any layer...

### Current (DOM-driven)
image

### Prototype
https://github.com/user-attachments/assets/95006924-913f-4025-a0ef-6d74b336ca67

### End Goal
image

Ref: https://github.com/joby-aviation/noodles.gl/discussions/50#discussioncomment-14799036

CC: @akre54 @ibgreen


examples/editable-layers/advanced/src/example.tsx
```diff
@@ -247,7 +247,7 @@ export default class Example extends React.Component<
       viewport: initialViewport,
       testFeatures: sampleGeoJson,
       mode: DrawPolygonMode,
-      modeConfig: null,
```
ibgreen: Let's clean up this PR so that it only contains changes to the infovis-layers module.

examples/infovis-layers/timeline/app.tsx
```diff
@@ -0,0 +1,734 @@
+// deck.gl-community
+// SPDX-License-Identifier: MIT
+// Copyright (c) vis.gl contributors
+
+import type {OrthographicViewState} from '@deck.gl/core';
```
ibgreen: This example  app.tsx file is very long, examples should be consise and easily readable to help users quickly understand the API. Could we break out some component like the timeline scrubber into a separate file?

examples/infovis-layers/timeline/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "jsx": "react-jsx"
```
ibgreen, line 4: Other examples don't need this. Try to follow the setup.

ibgreen:
> Do you have any snippets that you've experimented with to assist with the initial build of this, or do you recommend I take what I have and package it up myself?

yes it is all there, in module/infovis-layers/src/layers/...

charlieforward9:
Heres where I was able to land: 

https://github.com/user-attachments/assets/62913676-f6dc-48ef-823c-7aa4a8094943

A lot of pending work, I would appreciate a review and some support with the interaction logic. Still feels like there's too many keyboard and mouse events external to the component that I need to understand how to abstract better.

PR #444, opened 2025-11-22
feat(website) Add screenshot for leaflet gallery example

ibgreen:
- Compared to the original script, it looks like the API has changed from LeafletLayer to DeckLayer.
- TBH, I find both confusing as both basemaps and deck have Layer as central concept.

maybe `LeafletDeckOverlay` `LeafletOverlay`, `DeckOverlay` ? It would fit better with the other deck basemap integrations.

Also with leaflet API.
- L.ImageOverlay — overlaying a static image on a map bounding box.  [Leaflet](https://leafletjs.com/examples/overlays/?utm_source=chatgpt.com)
- L.VideoOverlay — overlaying a video element on a map. 
jgengle.github.io
- L.SVGOverlay — overlaying an SVG element on a map. (Mentioned in the Overlays docs) 
[Leaflet](https://leafletjs.com/examples/overlays/?utm_source=chatgpt.com)

### DIFF
--- a/examples/gallery/leaflet.html
+++ b/examples/gallery/leaflet.html
@@ -27,7 +27,6 @@
   <div id="map"></div>
 
   <script>
-    debugger
     const mapElement = document.querySelector('#map');
 
     const map = L.map(mapElement, {


zakjan:
Other base maps integrations use:
- bing-maps use `DeckOverlay`
- arcgis use `DeckLayer`

This is why I have done the original change from `LeafletLayer` to `DeckLayer`. The pattern to follow is `DeckXxx`.

Re whether `XxxLayer` or `XxxOverlay`, it's `DeckLayer` at the moment because it inherits from `L.Layer`. But now I see there are other `XxxOverlay` which inherit from `L.Layer` as well. It seems the difference is that `XxxLayer` are for tiled layers while `XxxOverlay` is more generic.

So yes, `DeckOverlay` might be the best after all. I'll submit the change.

PR #447, opened 2025-11-22
[editable-layers] Update to turf 7 and use geojson types

timnyborg:
Rebased and updated https://github.com/visgl/deck.gl-community/pull/221

### DIFF
--- a/modules/editable-layers/src/utils/utils.ts
+++ b/modules/editable-layers/src/utils/utils.ts
@@ -7,7 +7,7 @@ import bearing from '@turf/bearing';
 import pointToLineDistance from '@turf/point-to-line-distance';
 import {point} from '@turf/helpers';
 import {WebMercatorViewport} from 'viewport-mercator-project';
-import {Feature, Feature, LineString, Point, Position} from './geojson-types';
+import {Feature, LineString, Point, Position} from './geojson-types';
 import {Viewport} from './types';
 
 // TODO edit-modes: delete and use edit-modes/utils instead
--- a/modules/editable-layers/test/edit-modes/types.ts
+++ b/modules/editable-layers/test/edit-modes/types.ts
@@ -9,7 +9,6 @@ import {
   LineString,
   Polygon,
   MultiPolygon,
-  Feature,
   Feature
 } from '../../src/utils/geojson-types';
 
@@ -43,17 +42,20 @@ let multiPolygon: MultiPolygon = {
 
 let pointFeature: Feature<Point> = {
   type: 'Feature',
-  geometry: point
+  geometry: point,
+  properties: {}
 };
 
 let lineStringFeature: Feature<LineString> = {
   type: 'Feature',
-  geometry: lineString
+  geometry: lineString,
+  properties: {}
 };
 
 let anyFeature: Feature = {
   type: 'Feature',
-  geometry: multiPolygon
+  geometry: multiPolygon,
+  properties: {}
 };
 
 if (Math.random() > 0.5) {


charlieforward9 (commented):
A hefty lift! Thank you for this. If we can resolve these points below, I think we can have this merged in by today.

Great to see all the `@ts-expect-errors` go away.

Im seeing a TON of `FeatureCollection<SingleGeometry>` instances, this would be better as a shorthanded, single export util coming from the same file these types are defined. Cleans up dependent files with a single export rather than two. Open to just about any name that is a shorthand of this.

Can you explain the removal of all the files in `modules/editable-layers/src/lib/` and the `curve-utils`? Are they not used anywhere else across the repo or exported? (cc. @ibgreen, @georgioskarnas)

Also seeing a test fail from the geometry refactor.

timnyborg:
> Can you explain the removal of all the files in `modules/editable-layers/src/lib/` and the `curve-utils`? Are they not used anywhere else across the repo or exported? (cc. @ibgreen, @georgioskarnas)

From the original PR (I should copy the description across ):
> Drops code from long, long dead lib folder, and an unused curve-utils file, rather than trying to update turf usages there

A bunch of stuff from lib was exported (see the removals in `index.ts`), but none of it is relied on in the rest of the package, or referred to anywhere in the repo or the old nebula.gl documentation.  As far as I can tell, it's all dead.

charlieforward9:
Excellent. Thank you for your time. Gonna pull a senior collaborator in here for a final check, and once they confirm, we'll see this finally land.

modules/editable-layers/package.json
```diff
@@ -36,38 +36,37 @@
...
+    "@turf/meta": "7.2.0",
+    "@turf/midpoint": "7.2.0",
+    "@turf/nearest-point-on-line": "7.2.0",
+    "@turf/point-to-line-distance": "7.2.0",
+    "@turf/polygon-to-line": "7.2.0",
+    "@turf/rewind": "7.2.0",
+    "@turf/transform-rotate": "7.2.0",
+    "@turf/transform-scale": "7.2.0",
+    "@turf/transform-translate": "7.2.0",
+    "@turf/union": "7.2.0",
```
timnyborg: Just spotted these are exact versions.  Will revert to ^

modules/editable-layers/src/edit-modes/immutable-feature-collection.ts
```diff
@@ -3,20 +3,19 @@
...
   MultiLineString,
   MultiPolygon,
   Position,
-  PolygonCoordinates
+  GeometryFeature,
 } from '../utils/geojson-types';
 
 export class ImmutableFeatureCollection {
-  featureCollection: FeatureCollection;
+  featureCollection: GeometryFeatureCollection;
```
ibgreen: should this be 
```suggestion
  featureCollection: Readonly<GeometryFeatureCollection>;
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
+export { Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, Position };
 
-export type PointCoordinates = Position;
-export type LineStringCoordinates = Position[];
-export type PolygonCoordinates = Position[][];
-export type MultiPointCoordinates = Position[];
-export type MultiLineStringCoordinates = Position[][];
-export type MultiPolygonCoordinates = Position[][][];
+export type SingleGeometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
+export type GeometryFeature = Feature<SingleGeometry>
```
ibgreen: The next thing I would like to see is alignment of derived types' names with this type's name.

```suggestion
export type SimpleGeometryFeature = Feature<SimpleGeometry>
```

One option could be to "lift" the `Simple` concept and just go for 

```suggestion
export type SimpleFeature = Feature<SimpleGeometry>
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
+export type GeometryFeatureCollection = FeatureCollection<SingleGeometry>
```
ibgreen:
```suggestion
export type SimpleGeometryFeatureCollection = FeatureCollection<SimpleGeometry>
```

```suggestion
export type SimpleFeatureCollection = FeatureCollection<SimpleGeometry>
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
-export {Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, Position};
+export { Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, Position };
 
-export type PointCoordinates = Position;
-export type LineStringCoordinates = Position[];
-export type PolygonCoordinates = Position[][];
-export type MultiPointCoordinates = Position[];
-export type MultiLineStringCoordinates = Position[][];
-export type MultiPolygonCoordinates = Position[][][];
+export type SingleGeometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
```
ibgreen: The key naming decision in this PR is this new `SingleGeometry` type. We need it, but:

- The current name is slightly non-ideal because: “Single” could imply that it excludes MultiPoint / MultiLineString / MultiPolygon — but it includes them.
- “Single” is not used by GeoJSON, OGC, or Turf.

Proposals:
- `SimpleGeometry` (OGC-friendly or community-familiar)
  - A nod to OGC Simple Features, where the same six primitive geometry types exist.
 - Used by: OpenLayers (ol/geom/SimpleGeometry is a base class for non-collection geometries).
 - Pro- Familiar to people who know OGC / OpenLayers.
- Pro - Semantically correct (“simple” == “not a geometry collection”).
- Cons - Less common in the GeoJSON ecosystem.

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
-export type AnyCoordinates =
-  | PointCoordinates
-  | LineStringCoordinates
-  | PolygonCoordinates
-  | MultiPointCoordinates
-  | MultiLineStringCoordinates
-  | MultiPolygonCoordinates;
-
-export type Geometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
+export type AnyCoordinates = SingleGeometry['coordinates']
```
ibgreen:
```suggestion
export type SimpleGeometryCoordinates = SimpleGeometry['coordinates']
```

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
 
 export type Polygonal = Polygon | MultiPolygon;
```
ibgreen: Unrelated but this should probably have been `PolygonGeometry`

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
 
-// Feature types
+export { Feature, FeatureCollection } from 'geojson';
 
 export type BoundingBoxArray = [number, number, number, number];
+export type FeatureOf<T extends SingleGeometry> = Feature<T>;
```
ibgreen: Do we need to add/export these? Could they be internal? 

Are there non-simple counterparts? At least call them `SimpleFeatureOf`, ...

modules/editable-layers/src/utils/geojson-types.ts
```diff
@@ -11,64 +11,26 @@ import type {
...
-
-export type Geometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
+export type AnyCoordinates = SingleGeometry['coordinates']
 
 export type Polygonal = Polygon | MultiPolygon;
 
-// Feature types
+export { Feature, FeatureCollection } from 'geojson';
 
 export type BoundingBoxArray = [number, number, number, number];
```
ibgreen: It would be great to keep the file in sections, so we don't mix Geometry defs, Feature defs, other type defs, so that a user can quickly scan all exports of a given category.

e.g.

```
// Geometry exports

// Feature exports

// Other type exports
```

ibgreen (approved):

modules/editable-layers/src/utils/geojson-types.ts
timnyborg: I like that suggestion - probably the best of all the bad alternatives for 'Not a GeometryCollection'

modules/editable-layers/src/utils/geojson-types.ts
timnyborg: It's a pointless type now, since it just wraps Feature's generic typing

modules/editable-layers/src/utils/geojson-types.ts
timnyborg: After doing some more types clean-up, I'm not sure there's much value in grouping in that fashion - we're left with very little: 4 Simple* types (best understood together, I'd argue), PolygonGeometry, and AnyGeoJson.

timnyborg: Plus a bunch of geojson re-exports I want to remove in a follow-up

PR #448, opened 2025-11-24
fix: update import path for stylesheet and improve code formatting

charlieforward9:
widgets looking decent, we need to add the modeConfig I implemented for holes and intersections. Also, not sure if its my machine but its extremely laggy.

Would be nice to make this the standard example by consolidating all the advanced example features into it with a clean, reusable widget icon for every one.

### DIFF
--- a/examples/editable-layers/widget/example.tsx
+++ b/examples/editable-layers/widget/example.tsx
@@ -23,7 +23,7 @@ import {
 import StaticMap from 'react-map-gl/maplibre';
 import type {FeatureCollection} from 'geojson';
 
-import '@deck.gl/widgets/dist/stylesheet.css';
+import '@deck.gl/widgets/stylesheet.css';
 
 export function getDefaultGeoJSON(): FeatureCollection {
   return {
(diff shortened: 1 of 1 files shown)


examples/editable-layers/widget/example.tsx
```diff
@@ -23,7 +23,7 @@ import {
 import StaticMap from 'react-map-gl/maplibre';
 import type {FeatureCollection} from 'geojson';
 
-import '@deck.gl/widgets/dist/stylesheet.css';
+import '@deck.gl/widgets/stylesheet.css';
```
ibgreen, line 26: This still doesn't work in my environment. It fails to resolve when I run the example.

charlieforward9: Running with `/dist/` failed on my end, strange...

ibgreen: Both imports (with and without dist) fail on my end. This is very frustrating, wonder where our setups differ.

ibgreen (approved):
Approving since changing the import doesn't affect my inability to build :)

examples/editable-layers/widget/example.tsx
charlieforward9: I always refresh work in here by running `yarn` at the root anytime I pull from main. Maybe removing node_modules and retrying that will sync us up?

PR #450, opened 2025-11-26
Chore: import geojson types directly, do not re-export them

timnyborg:
Cleans up imports of geojson types, so all files import from `geojson` directly.

Removes export of standard geojson types from the library.  This _will_ require anyone doing an `import { FeatureCollection, ... } from '@deck.gl-community/editable-layers'` to follow suit.

### DIFF
--- a/modules/editable-layers/src/edit-modes/composite-mode.ts
+++ b/modules/editable-layers/src/edit-modes/composite-mode.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 // Copyright (c) vis.gl contributors
 
-import {FeatureCollection} from '../utils/geojson-types';
+import {FeatureCollection} from 'geojson';
 import {
   ModeProps,
   ClickEvent,
--- a/modules/editable-layers/src/edit-modes/delete-mode.ts
+++ b/modules/editable-layers/src/edit-modes/delete-mode.ts
@@ -1,4 +1,4 @@
-import {FeatureCollection} from '../utils/geojson-types';
+import {FeatureCollection} from 'geojson';
 
 import {GeoJsonEditMode} from './geojson-edit-mode';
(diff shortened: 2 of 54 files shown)


ibgreen:
Apart from the breaking change concern, I am just providing some perspective in case it could be helpful:

As we go down the track of reviewing the types in editable-layers, I think it could be worthwhile to discuss whether we keep the `geojson` dependency at all... 
- As a principle. in vis.gl we want our libraries to have as few dependencies as possible.
- One characteristic of TypeScript is that types are compatible if they are structurally equivalent
- This means that we don't necessarily need to import another library here, we could just define any types we need in geojson-types.ts.
- We'd just make sure that editable-layers would still be compatible with GeoJSON types the user might define, whether from `geojson` or other sources.

Overall, I feel that our focus should be on making the editable-layers API fully typed.
- My hunch is that giving the typing of the properties field more attention might be worthwhile. 
- I have rarely had problems with the non-properties GeoJSON types. 

Another thing we might possibly want to offer is a small zod schema for geojson types to help validate that input actually matches the types.
- One option could be to create a small `@math.gl/geojson` module with such features, then we'd keep this all within the vis.gl ecosystem and we could reuse such types in the deck.gl geojson layer etc.

timnyborg:
> @timnyborg Any thoughts, maybe if you can articulate why you suggested this change we can get to a consensus?

Honestly, just to clean up imports & exports that were made redundant by my turf 7 upgrade.  Turf vs nebula vs geojson type differences gave me no end of trouble over the years, so I see value in settling on the dominant library (geojson) and not reinventing the wheel.  That said, I probably won't be using these libraries again for the foreseeable 😢.  I've been ensuring all the fixes I submitted to nebula.gl over the years make it in, and they finally have, so I'm not bothered either way.

PR #454, opened 2025-11-30
Rename Leaflet DeckLayer to DeckOverlay

zakjan:
Ref: https://github.com/visgl/deck.gl-community/pull/444


ibgreen (approved):
Does the website gallery example still work? I was expecting that we'd need / want to use single file HTML+JS for each script example there? I.e the concern is around splitting into HTML + JS

PR #471, opened 2026-01-23
fix(editable-layers): use turf/kinks for self-inteserction test in DrawPolygonMode

### LINKED ISSUE #468: [Bug] DrawPolygonMode reports bogus intersection on valid polygons
### Module

- [ ] deck.gl-community/react

### Description

Example polygon:

```
{
    "type": "Feature",
    "properties": {},
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [ -92.44456305600066, 63.72868773989918 ],
                [ 44.23891326731302, 37.423106625872734 ],
                [ 43.63889683363783, 65.33932551389408 ],
                [ -92.44456305600066, 63.72868773989918 ]
            ]
        ]
    }
}
```

The `lineInteresect(...)` [here](https://github.com/visgl/deck.gl-community/blob/e7edaca05d5df052af45bcb215eec5ab4dceae75/modules/editable-layers/src/edit-modes/draw-polygon-mode.ts#L216) returns following intersection for this one:

```
[ 44.23891326731304, 37.423106625872734]
```

Note
1) it's not first point, but "second"
2) there is minor precision shift `44.23891326731304` vs `44.23891326731302`

Problem is easier to reproduce for large scale polygons - on world scale it's easiest.

Seems like we cannot assume that `lineIntersect` always finds intersection in first point and also comparison has to be done with some epsilon.

Note: In some related project, we're using `@turf/kinks` to check for polygon self-intersection, but honestly i'm not sure if this is same operation.

Workaround exists: disable self-intersection check using  `modeConfig.allowSelfIntersection = false` in `EditableGeoJsonLayer`

### Expected Behavior

`EditableGeoJsonLayer` in polygon mode with `allowSelfIntersection` enabled allow drawing non-intersecting polygons. 

### Steps to Reproduce

Go to https://visgl.github.io/deck.gl-community/examples/editable-layers/editor

On zoom level 0/1 Try to draw triangle that spans from North to South america, like on image below. When you "commit" polygon disappears.

Image

### Environment

- Framework version: deck.gl@9.2.6 / deck.gl-community/editable-layers@^9.2.0-beta.6
- Browser: Brave / Chrome
- OS: MacOS

### Logs

_No response_

zbigg:
Fix #468

* `editable-layers`: use `@turf/kinks` for self-inteserction test in DrawPolygonMode

### DIFF
--- a/modules/editable-layers/test/edit-modes/lib/draw-polygon-mode.spec.ts
+++ b/modules/editable-layers/test/edit-modes/lib/draw-polygon-mode.spec.ts
@@ -2,9 +2,9 @@
 // SPDX-License-Identifier: MIT
 // Copyright (c) vis.gl contributors
 
-import {beforeEach, describe, it, expect} from 'vitest';
-import {DrawPolygonMode} from '../../../src/edit-modes/draw-polygon-mode';
-import {createFeatureCollectionProps, createClickEvent, createKeyboardEvent, createPolygonFeature} from '../test-utils';
+import { beforeEach, describe, it, expect } from 'vitest';
+import { DrawPolygonMode } from '../../../src/edit-modes/draw-polygon-mode';
+import { createFeatureCollectionProps, createClickEvent, createKeyboardEvent, createPolygonFeature } from '../test-utils';
 
 let props;
 let mode;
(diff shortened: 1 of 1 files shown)


modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
```diff
@@ -213,18 +214,8 @@ export class DrawPolygonMode extends GeoJsonEditMode {
...
-      ).features.filter(
-        (intersection) =>
-          !newPolygon.geometry.coordinates[0].some(
-            (coord) =>
-              coord[0] === intersection.geometry.coordinates[0] &&
-              coord[1] === intersection.geometry.coordinates[1],
-          ),
-      );
-  
+      const overlapping = kinks(newPolygon).features;
```
charlieforward9, line 217: a much cleaner solution as well!

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
charlieforward9, line 1: a lot of formatting changes, was this file originally pushed without being linted or is this the way we want the file linted from now on? @ibgreen 

Id say I prefer the latter, but we should probably do this repo-wide rather than one file at a time.

Either way, I'd say this fix could land instead of blocking it for a formatting refactor

zbigg: Yes, sorry for this. Have habit of formatting and didn't noticed in pre-review - i can revert those.

(and yes, seems like original code wasn't formatted accordint to prettier here - not sure if this is expected or only preferred)

charlieforward9:
Just made a linting standard PR - with that and my publication out of the way - we will merge this one in and consider it the last unlinted piece of work in deck.gl-community. Thank you.

Issue #462, opened 2026-01-12
ThreeClickPolygon can error when using touch screen as lastPointerMoveEvent may be initially undefined


lottie-korf:
### Module

- [ ] deck.gl-community/arrow-layers
- [ ] deck.gl-community/bing-maps
- [x] deck.gl-community/editable-layers
- [ ] deck.gl-community/experimental
- [ ] deck.gl-community/graph-layers
- [ ] deck.gl-community/layers
- [ ] deck.gl-community/react
- [ ] deck.gl-community/react-graph-layer

### Description

The getGuides function is not defensive about lastPointerMoveEvent being undefined when operating on a touch screen device.
 
The touch event occurs before the mouse move event is set and this causes the error seen below:
<img width="1582" height="261" alt="Image" src="https://github.com/user-attachments/assets/9996bed6-eab6-4b05-ba3f-c03183ef443a" />

Attached is a screen recording of the issue.  You can see that an error is thrown on the initial click on the touchscreen.

https://github.com/user-attachments/assets/f1429922-82c8-422f-b4ff-19dd3809b4e4

This line in getGuides is what I think is causing the problem:

```JavaScript
  getGuides(props: ModeProps<FeatureCollection>): GuideFeatureCollection {
    const {lastPointerMoveEvent, modeConfig} = props;
    const clickSequence = this.getClickSequence();
 
    const guides: GuideFeatureCollection = {
      type: 'FeatureCollection',
      features: []
    };
 
    if (clickSequence.length === 0) {
      // nothing to do yet
      return guides;
    }
 
    // lastPointerMoveEvent may be undefined here because there is no pointer move event when using a touch screen
    const hoveredCoord = lastPointerMoveEvent.mapCoords;
	....
  }
```
I think other modes ARE defensive to this, such as draw-90degree-polygon-mode which does this:
```JavaScript
 getGuides(props: ModeProps<FeatureCollection>): GuideFeatureCollection {
    const guides: GuideFeatureCollection = {
      type: 'FeatureCollection',
      features: []
    };

    const clickSequence = this.getClickSequence();

    if (clickSequence.length === 0 || !props.lastPointerMoveEvent) {
      return guides;
    }
	...
	}
```

This means that I believe that the following fix will be appropriate:
```JavaScript
  getGuides(props: ModeProps<FeatureCollection>): GuideFeatureCollection {
    const { lastPointerMoveEvent } = props;
    const clickSequence = this.getClickSequence();

[...]

PR #482, opened 2026-02-12
Fix three click polygon mode, don't process guides unless we have a pointer move event

rjwats:
Possible fix for #462 which breaks three-click-polygon-mode on touchscreen devices.

There doesn't seem to be any test coverage for three-click-polygon-mode presently, however this fix seems to be present in several getGuides functions across the project.

### DIFF
--- a/modules/editable-layers/src/edit-modes/three-click-polygon-mode.ts
+++ b/modules/editable-layers/src/edit-modes/three-click-polygon-mode.ts
@@ -9,23 +9,29 @@ import {
   GuideFeatureCollection,
   TentativeFeature
 } from './types';
-import { Position, Polygon, Feature, FeatureCollection, SimpleFeatureCollection } from '../utils/geojson-types';
-import { GeoJsonEditMode } from './geojson-edit-mode';
+import {
+  Position,
+  Polygon,
+  Feature,
+  FeatureCollection,
+  SimpleFeatureCollection
+} from '../utils/geojson-types';
+import {GeoJsonEditMode} from './geojson-edit-mode';
 
 export class ThreeClickPolygonMode extends GeoJsonEditMode {
   handleClick(event: ClickEvent, props: ModeProps<SimpleFeatureCollection>) {
     this.addClickSequence(event);
 
-    const { modeConfig } = props;
+    const {modeConfig} = props;
     const clickSequence = this.getClickSequence();
 
     if (clickSequence.length > 2) {
-      const { geometry, properties } = this.getThreeClickPolygon(
+      const {geometry, properties} = this.getThreeClickPolygon(
         clickSequence[0],
         clickSequence[1],
         clickSequence[2],
         modeConfig
-      )
+      );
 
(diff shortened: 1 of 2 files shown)


rjwats:
Hi @charlieforward9,

Thanks for looking into this 👍

The cause of this issue is that there is that on most touchscreen devices there are no pointer move events, just touch events. The getGuides function is expecting there to have been a pointer move event before the the second touch event is received, but there won't have been, so this:

https://github.com/visgl/deck.gl-community/blob/f660c421b0e3779ae7abe99e61f21388b9d1cea9/modules/editable-layers/src/edit-modes/three-click-polygon-mode.ts#L54

errors with:

image

The fix we've implemented locally mirrors logic in other modes that caters for this condition:

https://github.com/visgl/deck.gl-community/blob/f660c421b0e3779ae7abe99e61f21388b9d1cea9/modules/editable-layers/src/edit-modes/two-click-polygon-mode.ts#L86

https://github.com/visgl/deck.gl-community/blob/f660c421b0e3779ae7abe99e61f21388b9d1cea9/modules/editable-layers/src/edit-modes/split-polygon-mode.ts#L84

Perhaps this issue was spotted and partially fixed before at some point :)

This issue can be reproduced in Chrome, even on the demo website:

- With chrome, visit https://visgl.github.io/deck.gl-community/examples/editable-layers/advanced
- Pull up developer tools
- Toggle on the device toolbar (which activates touch screen emulation - you no longer have a mouse pointer and Chrome emulates touch events when you click)
  image
- Turn on "Draw Rectangle Using 3 Points"
- Try and draw :)

image

I think was mostly concerned about having a regression check in the tests, but that test coverage isn't there at the moment.

charlieforward9:
thanks - looking forward to it. been having some challenges on mobile interactions myself id like to look into once I clean up this repo a bit more.

modules/editable-layers/src/edit-modes/three-click-polygon-mode.ts
```diff
@@ -46,7 +46,7 @@ export class ThreeClickPolygonMode extends GeoJsonEditMode {
       features: []
     };
 
-    if (clickSequence.length === 0) {
+    if (clickSequence.length === 0 || !lastPointerMoveEvent) {
```
ibgreen: Adding some comment about the reasons here would be great. Without that this could be hard to maintain.

charlieforward9:
Nice hefty analysis here @rjwats - thanks for digging in.

> I'm not sure why these would be coming back in a different order from Deck.gl's event handler does anyone know if this is expected or documented somewhere? I'm just concerned that this might be a workaround for an issue which should be fixed elsewhere.

https://github.com/visgl/mjolnir.js is exactly that event handling spec. May be insightful to look into here for additional context if you haven't poked around already.

One of the first issues that brought me into this community was related to [an unpropogated change from mjolnir](https://github.com/visgl/deck.gl-community/issues/201) - so it could be that this same issue is unsolved in the three-click-polygon-mode and just needs to be passed down.

PR #486, opened 2026-02-16
chore: enable repo-wide prettier formatting

charlieforward9:
## Summary

- Update `.prettierignore` to cover all modules (previously only graph-layers was formatted)
- Clean up `.eslintignore` to match
- Apply prettier repo-wide - mechanical formatting pass, no logic changes (82 files)

Addresses the formatting inconsistency discussed in #471 - contributors editors auto-formatting unformatted files was creating noisy diffs mixing style with code changes.

## Split PRs

Per review feedback, the following were split into separate PRs:
- Pre-commit hook: #492
- AGENTS.md code style docs: #493

## Commits

1. `chore: enable repo-wide prettier formatting` - config changes (.prettierignore, .eslintignore)
2. `chore: apply prettier formatting repo-wide` - mechanical formatting pass

## Test plan

- [x] All 248 tests pass (`yarn test`)
- [ ] `yarn lint` passes
- [ ] `yarn build` succeeds

PR #492, opened 2026-02-16
chore: enable pre-commit lint hook

charlieforward9:
## Summary

Wire up pre-commit hook via the already-installed `pre-commit` package. `yarn lint` now runs automatically before every commit.

Split out from #486 per review feedback - this needs a compatibility check with the publishing process before merging.

## Notes

- The `pre-commit` package is already in `devDependencies`
- This adds a single `pre-commit` config entry to `package.json`
- May need to verify this doesn't interfere with `ocular-publish`

## Test plan

- [ ] Verify `yarn lint` runs on `git commit`
- [ ] Verify `ocular-publish` still works with the hook enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,5 +76,9 @@
       }
     }
   },
+  "pre-commit": [
+    "lint",
+    "test"
+  ],
   "packageManager": "yarn@4.10.3"
 }


charlieforward9:
Updated this PR:
- Rebased on latest master (picks up vite, vitest, dev-tools bumps)
- Added `test` to the pre-commit hook alongside `lint`, matching the convention across other visgl repos (deck.gl, luma.gl, math.gl, react-map-gl all use `pre-commit` with similar config)
- Note: there are 3 pre-existing lint errors on master. The `viewport-mercator-project` one is fixed by #502. The experimental module ones need separate attention.
- Closed #501 (simple-git-hooks replacement) in favor of staying consistent with the visgl ecosystem.

PR #505, opened 2026-02-17
chore: remove dead code and outdated WIP content

charlieforward9:
## Summary
- Remove viv-layers fork (superseded by upstream @hms-dbmi/vividly)
- Remove basemap-layers from wip/ (merged into modules/basemap-props-wip in #504)
- Remove graph-viewer-legacy example (replaced by examples/graph-layers/graph-viewer)
- Remove stale codesandbox examples
- Remove implemented editable-layers RFCs

Supersedes #489. See also #504 for WIP module promotions.

## Test plan
- [x] All vitest tests pass (248 passed)
- [x] Pre-commit hooks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #496, opened 2026-02-16
RFC: Non-geospatial coordinate support for editable-layers


charlieforward9:
## Summary

Enable `editable-layers` to work with non-geospatial (Cartesian/screen-space) coordinate systems, not just WGS84 geographic coordinates.

From the [v9.3 tracker (#38)](https://github.com/visgl/deck.gl-community/issues/38) wishlist:
> - Goal: allow editable layers to be used with non-geospatial coordinates
> - Goal: make this a generic layer pack (non-geo-specific)

Originally described in the [2018 Generic EditMode RFC](https://github.com/visgl/deck.gl-community/tree/a6ec81d571ee2224291e83a0d0dc0419297b2c0b/dev-docs/RFCs/editable-layers/generic-edit-mode.md), which proposed `EditMode<TData, TGuides>` as a generic interface. The interface was implemented, but every concrete edit mode assumes geographic coordinates via turf.js.

## Current State

The `EditMode` interface is generic in TypeScript, but all implementations are geographic:

- **25 `@turf/*` packages** used across **43 files** — all assume WGS84 coordinates
- **Hardcoded units**: `kilometers` in circle/diameter/distance modes, `meters`/`centimeters` in split-polygon
- **Mercator projection**: `WebMercatorViewport` from legacy `viewport-mercator-project` package in `edit-modes/utils.ts`
- **Latitude-based math**: `metersPerPixel` formula in `elevation-handler.ts`
- **Rhumb line calculations**: `translate-from-center.ts` uses `@turf/rhumb-*`

The [`no-map` example](https://github.com/visgl/deck.gl-community/tree/master/examples/editable-layers/no-map) demonstrates a workaround — editing at the equator where Mercator distortion is minimal — but this is not real non-geo support.

### Geographic dependency by mode category

| Category | Modes | Turf dependency |
|---|---|---|
| Simple draw | DrawPoint, DrawLineString, DrawPolygon, DrawRectangle | Low — click-sequence math, but polygon validation uses `kinks`, `boolean-within`, `line-intersect` |
| Circle/ellipse | DrawCircleFromCenter, DrawCircleByDiameter, DrawEllipseByBoundingBox | High — `distance` + `circle` + `ellipse` with hardcoded `kilometers` |
| Transform | Modify, Translate, Rotate, Scale | Medium — `nearestPointOnProjectedLine()` uses Mercator, translate uses rhumb lines |
[...]

PR #512, opened 2026-02-18
docs(modules/nebula.gl)!: restore to glory

charlieforward9:
## Summary

- Adds `docs/modules/nebula.gl/design-rfc.md` preserving architectural rationale from the 4 original nebula.gl RFCs that were removed in #505
- Fixes pre-existing prettier formatting drift in 7 files that was blocking the pre-commit hook

## Context

The `dev-docs/RFCs/editable-layers/` directory was deleted in #505 with the message "implemented and no longer needed." While the *interfaces* are implemented, the RFCs contained valuable **design rationale** -- the "why" behind the architecture -- that was not captured anywhere in the current docs.

Key content preserved:
- **Generic EditMode RFC** -- why `ModeHandler` was replaced, the framework-agnostic design intent, module boundary rationale (edit-modes has no deck.gl dependency), and the non-geospatial extensibility goal
- **Tentative Feature RFC** -- the geometry type volatility problem and how separating tentative state from committed data solves it
- **Multi-Geometry Editing RFC** -- the unimplemented problem statement for appending to `Multi*` features
- **react-map-gl-draw RFC** -- the stateless component design, feature state machine, and why it was not ported

This is directly relevant to #496 (non-geospatial coordinate support), which references the Generic EditMode RFC by URL at the now-deleted commit.

## Test plan

- [ ] Verify the new doc renders correctly in the docusaurus sidebar
- [ ] Confirm prettier formatting changes are cosmetic only (arrow function parens, JSX line wrapping)

Relates to #496

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/modules/editable-layers/sidebar.json
+++ b/docs/modules/editable-layers/sidebar.json
@@ -36,7 +36,9 @@
     {
       "type": "category",
       "label": "Shader Modules",
-      "items": ["modules/editable-layers/api-reference/shader-modules/outline"]
+      "items": [
+        "modules/editable-layers/api-reference/shader-modules/outline"
+      ]
     }
   ]
 }


docs/modules/editable-layers/sidebar.json
```diff
@@ -7,6 +7,7 @@
     "modules/editable-layers/developer-guide/data-model",
     "modules/editable-layers/developer-guide/geojson-types",
     "modules/editable-layers/developer-guide/configuration",
+    "modules/editable-layers/developer-guide/design-decisions",
```
charlieforward9:
```suggestion
```

docs/modules/editable-layers/sidebar.json
```diff
@@ -36,9 +36,7 @@
     {
       "type": "category",
       "label": "Shader Modules",
-      "items": [
-        "modules/editable-layers/api-reference/shader-modules/outline"
-      ]
+      "items": ["modules/editable-layers/api-reference/shader-modules/outline"]
```
charlieforward9:
```suggestion
      "items": [
        "modules/editable-layers/api-reference/shader-modules/outline"
      ]
```

modules/editable-layers/src/widgets/editor-toolbar-widget.tsx
ibgreen: We don't normally use `Array` in typing

ibgreen:
Firstly, this seems to contain both a restoration of a removed file and a bunch of code changes. Should it be split into two PRs?

Just noting that the strong push you are driving to minimize directories in the root, while well-intended and valuable, is bringing up various issues: the directories that are being removed were there to avoid some practical issues / real concerns.

### On the RFC in docs folder topic:

- We have tried a number of different approaches to RFCs. Some RFCs have indeeed been raised as issues. These have the advantage that commenting on them is easier, but they do tend to get lost as time goes on. 
- I can't think of any obvious solution for RFCs that makes everyone comfortable, but at least let's keep the old ones in the source tree for now.
- If you really want the RFC in docs folder, would be good if there was a way to exclude them from search, but that requires more config etc.

### General comment

At this point, I think the best compromise might be to accept **one extra folder** in the root (perhaps called `dev`) and put all the stuff that is not part of the latest production builds there.
- RFCs
- wip modules
- wip examples
- ...

Mixing in outdated and half-finished files in the main directory structure will always raise concerns, some folks have very strong objections to including anything that is unfinished or confusing.

PR #532, opened 2026-02-25
refactor: remove preact dependency from widgets module

charlieforward9:
Replace all preact JSX/rendering with vanilla DOM APIs. The module
no longer requires preact as a dependency, fixing build failures
when consumed via Vite source aliases without preact installed.

- Convert 6 .tsx files to .ts (no more JSX)
- Replace preact Component/render/cloneElement/toChildArray/Fragment
  with plain DOM element creation and manipulation
- Introduce OverlayItemData interface (coordinates + createElement callback)
  as the framework-agnostic replacement for VNode
- Delete _deprecate/ directory (unused copies of widgets)
- Remove jsxImportSource: "preact" from tsconfig.json

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


ibgreen:
This would need to be raised with the TSC.
- Use of preact was a major decision in @chrisgervang 's design of the deck.gl/widgets module.
- The benefits in terms of readability etc were considered significant, and there are other benefits.
- The preact module is considered small and it is only used internally.
- Any widget that is intended to be upstreamed to deck.gl eventually (most of them) would need to be reconverted to preact before being accepted.

chrisgervang:
I can appreciate the build tool challenge preact may have caused, but I generally don't think that's a good reason to remove the dependency. I'd offer to help fix the broken tooling - where was it causing trouble?

PR #537, opened 2026-02-28
feat(editable-layers): non-geospatial coordinate system

eKerney:
Related to [RFC: Non-geospatial coordinate support for editable-layers #496](https://github.com/visgl/deck.gl-community/issues/496) 
Discussion here at: [feat(editable-layers): EditModeCoordinateSystem abstraction aligned with deck.gl COORDINATE_SYSTEM #516 ]( https://github.com/visgl/deck.gl-community/pull/516) 

Replaced TranslateMode with TranslateModeEx from @xinaesthete thank you!   
Added OrthographicView, adjusted background image bounds and INITIAL_VIEW_STATE accordingly.   
Added `COORDINATE_SYSTEM.CARTESIAN` to background and editableLayers.   
Changed selected feature color to make more obvious to user.   

Note: Running `yarn start-local` resulted in `Failed to resolve import` error, resolved by commenting line 22 in `vite.config.local.mjs`
```javascript
'@deck.gl': join(rootDir, './node_modules/@deck.gl'),
```

https://github.com/user-attachments/assets/3ce907a0-5265-4b13-aebb-31c65305830e

### DIFF
--- a/modules/editable-layers/src/edit-modes/utils.ts
+++ b/modules/editable-layers/src/edit-modes/utils.ts
@@ -130,7 +130,7 @@ export function projectOrUnprojectPoints(
   project: 'PROJECT' | 'UNPROJECT',
   wmViewport?: WebMercatorViewport
 ): Position {
-  if (coordinateSystem === undefined) return [...inputPoints];
+  if (coordinateSystem === undefined || wmViewport === undefined) return [...inputPoints];
   else if (coordinateSystem instanceof CartesianCoordinateSystem) return [...inputPoints];
   return project === 'PROJECT'
     ? wmViewport.project([...inputPoints])
@@ -227,10 +227,7 @@ export function nearestPointOnLine(
   viewport?: Viewport,
   coordinateSystem?: EditModeCoordinateSystem
 ): NearestPointType {
-  const viewportCheck = viewport
-    ? viewport
-    : {width: 0, height: 0, longitude: 0, latitude: 0, zoom: 0};
-  const wmViewport = new WebMercatorViewport(viewportCheck);
+  const wmViewport = viewport ? new WebMercatorViewport(viewport) : undefined;
 
   let closestPoint: any = point([Infinity, Infinity], {dist: Infinity});
 


charlieforward9:
@eKerney you could try putting an agent on it and directing it towards the feedback

eKerney:
@charlieforward9 made a very small change by swapping `TranslateModeEx` for the `editable-layers/TranslateMode` and everything works great!  editableLayer coordinateSystem: `COORDINATE_SYSTEM.CARTESIAN` - from deck Core.

charlieforward9 (changes requested):
Sweet, simple change! 

Id say the real work with this coordinate task is to pull the edit mode widget into the no map example and show all the relevant modes with non-geographic values. 

Basically have the editable layer operations be controlled by the active deck coordinate system

* if geo - work as is
* if non geo - handle that maths without turf.js.

Non non geo situation could have editable layers pulling in way less dependencies

modules/editable-layers/src/edit-modes/utils.ts
```diff
@@ -184,19 +221,16 @@ export function nearestPointOnProjectedLine(
...
-  if (viewport) {
-    mercator = new WebMercatorViewport(viewport);
-  }
-  let closestPoint: any = point([Infinity, Infinity], {
-    dist: Infinity
-  });
+  let closestPoint: any = point([Infinity, Infinity], {dist: Infinity});
 
   if (!lines.geometry?.coordinates.length || lines.geometry?.coordinates.length < 2) {
     return closestPoint;
```
charlieforward9, line 235: `new WebMercatorViewport(viewport)` is now unconditional but `viewport` can be `undefined`. Old code guarded this with `if (viewport)`. In cartesian-only usage (no map) this will throw. Need the guard back or make viewport optional and skip projection when its absent.

modules/editable-layers/src/edit-modes/utils.ts
```diff
@@ -116,20 +124,49 @@ export function mix(a: number, b: number, ratio: number): number {
...
 
+export function projectOrUnprojectPoints(
+  inputPoints: Position,
+  coordinateSystem: EditModeCoordinateSystem,
+  wmViewport: WebMercatorViewport,
+  project: 'PROJECT' | 'UNPROJECT'
+): Position {
+  if (coordinateSystem === undefined) return [...inputPoints];
+  else if (coordinateSystem instanceof CartesianCoordinateSystem) return [...inputPoints];
+  return project === 'PROJECT'
```
charlieforward9, line 135: Consider making `wmViewport` optional here. Cartesian path returns early without using it, but callers still have to construct one just to pass it in. Would also clean up the viewport guard issue in `nearestPointOnLine`.

modules/editable-layers/src/edit-modes/draw-line-string-mode.ts
```diff
@@ -37,11 +41,11 @@ export class DrawLineStringMode extends GeoJsonEditMode {
       this.addClickSequence(event);
       positionAdded = true;
     }
-    const clickSequence = this.getClickSequence();
+    const clickSequence: Position[] = this.getClickSequence();
```
charlieforward9, line 44: What happens when `props?.coordinateSystem` is undefined? Gets passed straight to `getEditModeCoordinateSystem()` - make sure that has a sensible default (fall back to geo) otherwise this breaks existing geo workflows that dont set it.

modules/editable-layers/test/edit-modes/lib/utils.spec.ts
```diff
@@ -249,4 +250,14 @@ describe('nearestPointOnLine()', () => {
     expect(result.geometry.coordinates[1]).toBeCloseTo(0.0, 3);
     expect(result.properties.index).toEqual(0);
   });
+
+  it('Correctly intersects line normal to slope of line WITH CARTESIAN COORDINATES ', () => {
+    // @ts-expect-error TODO
+    const result = nearestPointOnLine(LineString, Point, viewport, COORDINATE_SYSTEM.CARTESIAN);
```
charlieforward9: The `@ts-expect-error` here means `COORDINATE_SYSTEM.CARTESIAN` (number enum from deck core) doesnt match the `EditModeCoordinateSystem` type. Should use `CartesianCoordinateSystem` from #516 so the test actually validates the type contract instead of bypassing it.

modules/editable-layers/test/edit-modes/lib/modify-mode.spec.ts
```diff
@@ -458,6 +458,17 @@ describe('getGuides()', () => {
...
         cancelPan: vi.fn(),
         sourceEvent: null
+      },
+      modeConfig: {
+        viewport: {
+          width: 800,
+          height: 600,
+          latitude: 37.8,
+          longiude: -122.4,
+          zoom: 10,
```
charlieforward9, line 468: Typo: `longiude` -> `longitude`

modules/editable-layers/src/edit-modes/draw-line-string-mode.ts
eKerney: Updated `calculateInfoDraw` and `_getTooltips` to instantiate a new GeoCoordinateSystem if undefined.

modules/editable-layers/src/edit-modes/utils.ts
eKerney: This was throwing an error for Geomode editing, most edit modes don't pass a viewport to nearestPointOnLine.  Not wanting to change the function signature since this is implemented in many places, I've added a rather inelegant viewportCheck which returns the viewport or a viewport placeholder.   No errors testing with no-map & editor examples

modules/editable-layers/test/edit-modes/lib/modify-mode.spec.ts
eKerney: done

modules/editable-layers/test/edit-modes/lib/utils.spec.ts
eKerney: Changed `COORDINATE_SYSTEM.CARTESIAN` to `new CartesianCoordinateSystem`.  
The original @ts-expect-error was due to improper type match for LineString, Point, and viewport.  Added empty properties and const string to match function signature and proper viewport placeholder resolving TS errors warnings in this test.

modules/editable-layers/src/edit-modes/utils.ts
eKerney: Made it optional and swapped places with project string literal binding, adjusted any calls in utils.ts to match updated signature.,

modules/editable-layers/src/edit-modes/utils.ts
charlieforward9: Now that `wmViewport` is optional in `projectOrUnprojectPoints` (nice fix btw), the placeholder object isn't needed anymore. Should be able to just do:

```ts
const wmViewport = viewport ? new WebMercatorViewport(viewport) : undefined;
```

and let the cartesian early-return in `projectOrUnprojectPoints` handle the rest.

eKerney: @charlieforward9 removed placeholder object, needed to add check for early return with `wmViewport === undefined` in `projectOrUnprojectPoints` - throwing an error in edit mode with editor demo.

Issue #596, opened 2026-04-17
[Feat] @deck.gl/json v2 Tracker


ibgreen:
### Target Use Case

- Enable LLMs to generate deck.gl visualizations.
- Enable JSON schemas for editors like Monaco
- Add zod definitions to validate deck.gl JSON.
- Strengthen pydeck capabilities
- Better playground application
- ...

### Proposal

Create a new module in deck.gl-community where we iterate and build a clean solution that we then propose for upstreaming into deck.gl/json

### Details

- [ ] Move the expression parser in deck.gl to `@math.gl/expressions`
- [ ] Implement a catalog of Zod schemas for deck layer props, views etc.
- [ ] Generate JSON schemas for deck.gl JSON from the Zod schemas and publish/export them.
- [ ] Build out a playground v2 example in community repo that leverages all the features.
- [ ] Make the JSON LLM friendly, examples docs LLM.txt etc so that LLMs can easily generated correct deck.gl JSON.
- [ ] Provide a way to hook data sources into deck.gl/json, especially SQL sources (e.g. see the new SQLSourceLoader in upcoming loaders.gl v5.0 release)
- [ ] ...

ilyabo:
I think it'd be good also to support [deck.gl-geoarrow](https://github.com/geoarrow/deck.gl-geoarrow) layers (and/or a straightforward way to extend the spec with new layers for clients) and a way to pass data (esp. arrow) separately from the layers spec.

charlieforward9:
@ilyabo tracking that in #79

ilyabo:
I also started some related work in SQLRooms: https://github.com/sqlrooms/sqlrooms/issues/548

dzole0311:
👋 sharing a few thoughts. I find the exisitng `@deck.gl/json` has a nice runtime conversion model where apps provide a catalog of classes, functions, constants and enums and then JSON uses that to resolve layers, views and functions.

I was going through the PR referenced above (https://github.com/visgl/deck.gl-community/pull/597) and I wonder if it would help to define the base schema architecture before adding many specific schemas, so if the base is clear, individual contributions become easier to review and easier to sequence later.

The pieces I'd imagine that base needing:

- a schema/catalog API for registering layer and view schemas
- shared schemas for common deck.gl concepts (base layer props, views, accessors, colors, data bindings)
- and ideally a split between the visualization spec and data-source declarations (eg. some consideration of how data-source declarations relate to the visualization spec especially for SQL-backed sources)

[...]

PR #597, opened 2026-04-17
feat: @deck.gl-community/json — Zod schemas for GeoJSON (Step 1 of #596)

charlieforward9:
## Summary

Related to #596.

This narrows the PR to the optional `@deck.gl-community/json` module only:

- adds shared Zod-backed GeoJSON schemas and inferred TypeScript types
- covers Position, BBox, all GeoJSON geometry primitives, Feature, and FeatureCollection
- wires the new workspace into TypeScript/Vitest resolution
- adds focused schema coverage for valid and invalid GeoJSON shapes
- removes the editable-layers AI-tools coupling from this PR so that work can be reviewed separately

## Validation

- `yarn`
- `yarn lint`
- `yarn lint-fix`
- `yarn test-node modules/json/src/test/geojson.test.ts`
- pre-commit `yarn test`: 164 files passed, 1423 tests passed, 9 skipped
- `yarn build`

## Manual verification

- No local preview URL is needed for this package-only/docs-note change.
- Website production build generated static files successfully.

## Notes

This keeps the module validator-specific for now because the existing PR and tests are built around Zod schemas, but it aligns with the repo's current Zod v4 dependency instead of adding a second Zod v3 range.


ibgreen:
My preference would be to land the new module with the schemas separately, and then open another PR with any changes you want to do in editable layers.

Generally, my view is that the json module should be an optional module that users import when they want to add a JSON support, we will likely add a lot of stuff to that module, other modules like editable layers should not import it.

For something like a GeoJSON zod schema, that could potentially go into a loaders.gl module.

organize:
just passing by, but surely this can be achieved without tying consumers to a specific zod version (or zod in general)? I'd like to understand the version selected here - I'm not sure how this will play out for anyone using v4 since a lot of deprecations are being used. strictly speaking, some of the `optional` semantics might be inaccurate (particularly for anyone using `exactOptionalPropertyTypes`).

while I understand this is an experimental package, I don't think it's a good idea to bake in a specific validator. why not export Standard Schema / Standard JSON Schema and let the consumers use a validator of their choice?

sorry if this sounds negative, it's not my intention. I truly appreciate the hard work you maintainers are putting in  🤝 ❤️

PR #611, opened 2026-04-26
chore: Replace eslint/prettier with biome

ibgreen:
## Summary

Replaces the repo-wide Prettier/ESLint lint setup with Biome, following the `loaders.gl` / `luma.gl` devtools-extension pattern.

## Changes

- Added root `biome.jsonc` with formatter and linter rules.
- Added `dev-modules/devtools-extensions` with:
  - shared Biome base config
  - `lint.mjs` wrapper
  - local `@biomejs/biome` dependency
- Updated root scripts:
  - `yarn lint`
  - `yarn lint-fix`
  - `yarn lint-pre-commit`
- Removed old ESLint/Prettier config files:
  - `.eslintrc.cjs`
  - `.eslintignore`
  - `.prettierrc`
  - `.prettierignore`
- Configured Biome warnings to fail lint via `--error-on-warnings`.
- Promoted unused import/variable checks to errors and cleaned up existing diagnostics.
- Ran Biome formatting across the JS/TS source tree.
- Updated contributor/agent docs to reference Biome instead of Prettier/ESLint.
- Updated `yarn.lock`.

## Verification

- `yarn`
- `yarn lint`
- `yarn build`
- `yarn test-node`

`yarn build` passes with existing build warnings about `import.meta` in CJS output / `NO_COLOR`, but exits successfully.


jake-goldsmith:
Hello,

After this PR I've been having issues running this on Windows - I don't know - does this project support development on Windows or is it expected to be on unix? I'm not an expert in yarn and how this is all set up but this was the feedback from the LLM about the issue:

> `yarn lint` and `yarn lint-fix` silently fail on Windows (exit code 1, no output).
> **Root cause:** `dev-modules/devtools-extensions/biome/lint.mjs` calls `spawnSync('yarn', ['exec', 'biome', ...])` with `cwd` set to the repo root. On Windows, spawned child processes don't have `node_modules/.bin` on their PATH unless `shell: true` is set, so `yarn exec biome` fails to resolve the binary even though it exists at `node_modules/.bin/biome`.
> **Fix:** call the biome binary directly by resolved path and add `shell: true`:

```js
const biomeBin = path.join(repoRoot, 'node_modules', '.bin', 'biome');
// ...
spawnSync(biomeBin, [command, ...args], {cwd: repoRoot, stdio: 'inherit', shell: true});
```

PR #629, opened 2026-05-07
fix(editable-layers) Update nearestPointOnLine so it saves distance correctly

### LINKED ISSUE #622: [Bug] Editable Layers - Modify mode intermediate handle renders in wrong place
### Module

- [ ] deck.gl-community/react

### Description

In modify mode the intermediate handle should be at the nearest point to the cursor. Sometimes if there are multiple lines as part of the feature it picks the wrong one.

I think the issue is in the nearestPointOnLine function - https://github.com/visgl/deck.gl-community/blob/a437ed183f03cdb54c2503cde0377e58de5ca68c/modules/editable-layers/src/edit-modes/utils.ts#L224-L351

It is comparing the distance as follows but I don't think that reflects the actual distance between points:

Image

### Expected Behavior

_No response_

### Steps to Reproduce

On the example at https://visgl.github.io/deck.gl-community/examples/editable-layers/advanced

If you select modify mode, select the polygon then some of the edges don't render the intermediate handle properly when the cursor is above them.

For example here my cursor is on the lower line but the intermediate handle renders on the higher one.

Image

There is also similar behaviour on the MultiLineString feature.

### Environment

- Browser: Chrome
- OS: Windows

### Logs

_No response_

jake-goldsmith:
Closes #622 

Simplifies the nearestPointOnLine function - also removes dead code for handling multi part geometries since they are handled at higher levels.

The key bugfix here is setting `dist: minDist` on the returned object. Previously it was returning a different distance metric so when handling features with multiple lines the results were being compared incorrectly. 

Modify Mode intermediate points are now working as expected:

https://github.com/user-attachments/assets/00ef1f86-4978-461e-988e-3f672f3a14f8

### DIFF
--- a/modules/editable-layers/src/edit-modes/utils.ts
+++ b/modules/editable-layers/src/edit-modes/utils.ts
@@ -228,11 +228,9 @@ export function nearestPointOnLine(
 ): NearestPointType {
   const wmViewport = viewport ? new WebMercatorViewport(viewport) : undefined;
 
-  const closestPoint = point([Infinity, Infinity], {dist: Infinity, index: undefined});
-
   const coords = lines.geometry?.coordinates;
   if (!coords || coords.length < 2) {
-    return closestPoint;
+    return point([Infinity, Infinity], {dist: Infinity, index: undefined});
   }
 
   const pointCoords = getCoords(inPoint);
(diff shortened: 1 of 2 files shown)


charlieforward9 (approved):
Im curious to see this tested on a GlobeView as well

according to gpt 

> it computes the nearest point in Web Mercator/screen-projected space, then unprojects the result. That is usually appropriate for “nearest visible point to the cursor” in an editor, but it is still not the same as shortest distance on the ellipsoid/sphere

something to eventually get around to later - thanks for the fix!

PR #635, opened 2026-05-13
fix(editable-layers): correct SelectionLayer polygon picking for elongated lassos

llamington:
## Summary

Fixes false-positive selection in `SelectionLayer`'s polygon mode for elongated/diagonal/concave lassos by replacing the GPU-pick + buffered-blocker mechanism with a CPU point-in-polygon test.

## The bug

When using `SelectionLayer({selectionType: 'polygon', ...})`, pins that sit *outside* the drawn lasso are returned in `onSelect({pickingInfos})` as if they were inside. The failure is reliable for any lasso whose bounding box extends more than ~50 km past its own edges — long thin lassos, diagonal lassos, and concave shapes.

### Repro

1. Render a `ScatterplotLayer` of pins spread across a continent (e.g. the existing Manhattan demo, but at a wider zoom).
2. Add a `SelectionLayer` with `selectionType: 'polygon'`.
3. Draw a long thin lasso diagonally across the map.
4. Inspect `onSelect({pickingInfos})`: pins clearly outside the drawn polygon are included.

## Root cause

`_selectPolygonObjects` does this:

1. Builds a "blocker" polygon — a `turfBuffer(lasso, EXPANSION_KM = 50)` polygon with the lasso subtracted (`turfDifference`).
2. Calls `deck.pickObjects({bbox: lassoBoundingBox, layerIds: [blockerId, ...layerIds]})` after a `setTimeout(250)` for the blocker to render.
3. Filters the picks.

The blocker is what's meant to mask out pins outside the polygon. But:

- The buffer is fixed at 50 km. Any lasso whose bbox extends >50 km past its edges has gaps where the blocker doesn't cover, and pins in those gaps get picked even though they're outside the lasso polygon.
- The `setTimeout(250)` is a race; the source acknowledges it: `// HACK, find a better way` (line 139).
- The picking framebuffer also drops overlapping pins at the same pixel (related issue, [nebula.gl#297](https://github.com/uber/nebula.gl/issues/297)).

[nebula.gl#132](https://github.com/uber/nebula.gl/issues/132) ("Fix concave lasso select bug") has been open since 2018 with a noted TODO: "more proper implementation than the setTimeout hack." This PR is that implementation.

## The fix

Replace the GPU-pick + blocker mechanism with a direct CPU point-in-polygon test:

1. Take the drawn polygon coordinates.
2. For every layer named in `layerIds`, iterate its `data`.
3. Resolve each datum's position via `layer.props.getPosition(...)`, with `object.position` / `object.coordinates` fallbacks.
4. Run `@turf/boolean-point-in-polygon` (already a dependency) against the polygon.
5. Emit a pickingInfo `{object, layer, index}` for each match.

The result is exact, synchronous, and free of the buffer, framebuffer, and timing constraints.

## Scope

The new implementation supports layers exposing per-datum positions — `ScatterplotLayer`, `IconLayer`, `TextLayer`, `ColumnLayer`, etc. This matches:

- The layer type in [the docs example](https://github.com/visgl/deck.gl-community/blob/master/docs/modules/editable-layers/api-reference/layers/selection-layer.md) (`ScatterplotLayer`).
- The layer type in `examples/editable-layers/sf/app.tsx` and `examples/editable-layers/advanced/src/example.tsx` (`EditableGeoJsonLayer`).

The previous implementation incidentally returned pickingInfos for `PathLayer` / `PolygonLayer` / `ArcLayer` via GPU picking, but the results were never meaningful — the 50 km blocker meant the lasso shape barely applied to those geometries anyway. Polygon-mode support for non-point layers was never documented, and no example demonstrates it. If maintainers want explicit support later, that's a separate feature with its own UX decisions (intersection vs containment for paths, etc.).

### DIFF
--- a/modules/editable-layers/src/editable-layers/selection-layer.ts
+++ b/modules/editable-layers/src/editable-layers/selection-layer.ts
@@ -4,17 +4,15 @@
 
 /* eslint-env browser */
 
-import type {CompositeLayerProps, DefaultProps} from '@deck.gl/core';
+import type {CompositeLayerProps, DefaultProps, Layer, PickingInfo} from '@deck.gl/core';
 import {CompositeLayer} from '@deck.gl/core';
-import {PolygonLayer} from '@deck.gl/layers';
-import {featureCollection, polygon} from '@turf/helpers';
-import turfBuffer from '@turf/buffer';
-import turfDifference from '@turf/difference';
+import booleanPointInPolygon from '@turf/boolean-point-in-polygon';
+import {point, polygon} from '@turf/helpers';
 
-import {EditableGeoJsonLayer} from './editable-geojson-layer';
-import {DrawRectangleMode} from '../edit-modes/draw-rectangle-mode';
 import {DrawPolygonMode} from '../edit-modes/draw-polygon-mode';
+import {DrawRectangleMode} from '../edit-modes/draw-rectangle-mode';
 import {ViewMode} from '../edit-modes/view-mode';
+import {EditableGeoJsonLayer} from './editable-geojson-layer';
 
 export const SELECTION_TYPE = {
   NONE: null,
@@ -49,9 +47,7 @@ const EMPTY_DATA = {
   features: []
 };
 
-const EXPANSION_KM = 50;
 const LAYER_ID_GEOJSON = 'selection-geojson';
-const LAYER_ID_BLOCKER = 'selection-blocker';
 
 const PASS_THROUGH_PROPS = [
   'lineWidthScale',
@@ -75,17 +71,12 @@ const PASS_THROUGH_PROPS = [
   'getTentativeFillColor',
   'getTentativeLineWidth'
 ];
+
 export class SelectionLayer<DataT, ExtraPropsT> extends CompositeLayer<
   ExtraPropsT & Required<SelectionLayerProps<DataT>>
 > {
   static layerName = 'SelectionLayer';
   static defaultProps = defaultProps;
-
-  state: {
-    pendingPolygonSelection: {
-      bigPolygon: ReturnType<typeof turfDifference>;
-    };
-  } = undefined!;
 
   _selectRectangleObjects(coordinates: any) {
     const {layerIds, onSelect} = this.props;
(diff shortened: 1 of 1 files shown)


modules/editable-layers/src/editable-layers/selection-layer.ts
```diff
@@ -104,57 +95,26 @@ export class SelectionLayer<DataT, ExtraPropsT> extends CompositeLayer<
...
+    const pickingInfos: SelectionPickingInfo[] = this.context.layerManager
+      .getLayers()
+      .filter(layer => layerIds.includes(layer.id))
+      .flatMap(layer => {
+        const data = layer.props.data;
+        if (!Array.isArray(data)) return [];
+        return data.flatMap((object, index): SelectionPickingInfo[] => {
+          const position = extractPosition(layer, object, index, data);
+          if (position === null) return [];
+          if (!booleanPointInPolygon(point(position), selectionPolygon)) return [];
```
charlieforward9, line 109: non-geo may not all pass this conditional check

This works better than the current approach, so its not blocking, but these instances of turf should be triaged under a more careful pass - which may have been the original intention of the GPU-based selections?

CC @eKerney

modules/editable-layers/src/editable-layers/selection-layer.ts
```diff
@@ -187,29 +147,41 @@ export class SelectionLayer<DataT, ExtraPropsT> extends CompositeLayer<
...
   }
 
   shouldUpdateState({changeFlags: {stateChanged, propsOrDataChanged}}: Record<string, any>) {
     return stateChanged || propsOrDataChanged;
   }
 }
+
+type SelectionPickingInfo = Pick<PickingInfo, 'object' | 'layer' | 'index'>;
+
+function extractPosition(
```
charlieforward9, line 159: Would we want to extract a 3D position here for any reason? @felixpalmer

charlieforward9:
I opened #674 as a focused Codex follow-up carrying the polygon lasso selection fix onto a repo branch with added SelectionLayer regression coverage and GitHub-visible SF Polygons proof media. I avoided pushing to this contributor branch directly; maintainers can either review #674 or use its tests/proof to update this PR.

PR #641, opened 2026-05-22
finish draw-polygon mode cartesian and add tests

eKerney:
Finished updating `draw-polygon-mode.ts` to use turf(geo modes), and simple geometric functions(cartesian modes).  Did not attempt turfJS kinks function which involves another order of complexity.  Created simple geometry functions and tests, needed to bypass linter checks for complexity on `validateAndCreateHole()`.

### DIFF
--- a/modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
+++ b/modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
@@ -305,14 +305,7 @@ export class DrawPolygonMode extends GeoJsonEditMode {
     }
 
     // Check outer polygon conflicts after iteration
-    // const intersectionWithOuter = cartesian
-    //   ? polygonEdgesIntersect(outer, newPolygon)
-    //   : lineIntersect(outer, newPolygon).features.length > 0;
-    const intersectionWithOuter = isIntersectionWithOuter(
-      outer,
-      newPolygon,
-      props.coordinateSystem
-    );
+    const intersectionWithOuter = isPolygonIntersecting(outer, newPolygon, props.coordinateSystem);
     if (intersectionWithOuter) {
       props.onEdit({
         updatedData: props.data,
@@ -322,9 +315,7 @@ export class DrawPolygonMode extends GeoJsonEditMode {
       return {handled: true};
     }
 
-    const outerWithin = cartesian
-      ? polygonWithinPolygon(outer, newPolygon)
-      : booleanWithin(outer, newPolygon);
+    const outerWithin = isPolygonWithin(outer, newPolygon, props.coordinateSystem);
     if (outerWithin) {
       props.onEdit({
         updatedData: props.data,
@@ -335,9 +326,7 @@ export class DrawPolygonMode extends GeoJsonEditMode {
     }
 
     // Check if new polygon is within outer polygon (valid hole)
-    const newWithinOuter = cartesian
-      ? polygonWithinPolygon(newPolygon, outer)
-      : booleanWithin(newPolygon, outer);
+    const newWithinOuter = isPolygonWithin(newPolygon, outer, props.coordinateSystem);
     if (newWithinOuter) {
       const updatedData = new ImmutableFeatureCollection(props.data)
         .replaceGeometry(featureIndex, {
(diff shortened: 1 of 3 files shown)


marcomuser:
Fantastic! I just stumbled over a similar issue with the ThreeClickPolygonMode. That would also need a cartesian handling. Are you planning to handle the other modes as well in future PRs?

I'm currently subclassing that mode to make it work with cartesian layers:
```ts
class CartesianDrawRectangleUsingThreePointsMode extends DrawRectangleUsingThreePointsMode {
  override getThreeClickPolygon(
    coord1: number[],
    coord2: number[],
    coord3: number[]
  ) {
    const [x1, y1] = coord1;
    const [x2, y2] = coord2;
    const [x3, y3] = coord3;

    const ex = x2 - x1;
    const ey = y2 - y1;
    const len = Math.sqrt(ex * ex + ey * ey);
    if (len === 0) return null;

    const px = -ey / len;
    const py = ex / len;
    const d = (x3 - x1) * px + (y3 - y1) * py;

    const p3 = [x2 + d * px, y2 + d * py];
    const p4 = [x1 + d * px, y1 + d * py];

    return {
      type: 'Feature' as const,
      properties: { shape: 'Rectangle' },
      geometry: {
        type: 'Polygon' as const,
        coordinates: [[coord1, coord2, p3, p4, coord1]],
      },
    };
  }
}
```

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
charlieforward9, line 1: This works - so a merge could happen and this could be tackled later - but my main concern is its deeply conditional within the mode logic. A better approach would be to initiate the coordiate-system-specific utilities in a seperate file and allow them to be reused across the modes. 

This would help remove the direct `turf` dependency in the mode definitions over time and seperate geometry math from mode logic.

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
```diff
@@ -262,21 +268,24 @@ export class DrawPolygonMode extends GeoJsonEditMode {
...
     // Check existing holes for conflicts
     for (let i = 1; i < feature.geometry.coordinates.length; i++) {
-      const hole = turfPolygon([feature.geometry.coordinates[i]]);
-      const intersection = lineIntersect(hole, newPolygon);
+      const hole = getPolygonFeature([feature.geometry.coordinates[i]], props);
 
-      if (intersection.features.length > 0) {
+      const intersection = cartesian
+        ? polygonEdgesIntersect(hole, newPolygon)
+        : lineIntersect(hole, newPolygon).features.length > 0;
```
charlieforward9:
```suggestion
      const intersection = isPolygonIntersecting(hole, newPolygon, coordinateSystem);
```

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
```diff
@@ -285,7 +294,10 @@ export class DrawPolygonMode extends GeoJsonEditMode {
         return {handled: true};
       }
 
-      if (booleanWithin(hole, newPolygon) || booleanWithin(newPolygon, hole)) {
+      const containsOrContained = cartesian
+        ? polygonWithinPolygon(hole, newPolygon) || polygonWithinPolygon(newPolygon, hole)
+        : booleanWithin(hole, newPolygon) || booleanWithin(newPolygon, hole);
```
charlieforward9:
```suggestion
      const containsOrContained = isContainingOrContained(poly1, poly2, coordinateSystem);
```

charlieforward9 (approved):

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
charlieforward9, line 1: I still feel like all the coordinate system conditionals should not live within the mode-logic. `getPolygonFeature` implementation is a good case for what I believe would look cleanest as we propogate this throughout all the modes.

charlieforward9 (changes requested):

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
eKerney: Applied suggestion, see below

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
eKerney: Applied suggestion, see below

modules/editable-layers/src/edit-modes/draw-polygon-mode.ts
eKerney: @charlieforward9 - I've addressed your comments regarding moving the coordinate system conditional logic away from the mode-logic, and created helper functions such as `isPolygonWithin()`.  In place of ~20 line repeated functions with switch/case, I've gone with a custom dispatch `forCoordSystem()` which reduces repeated switch logic and is easily extendable.  Also created a cartesian-utils.ts file to separate the new non-turf functionality more cleanly without the turf dependency.  All tests are go :)

eKerney:
@marcomuser - I've taken a crack at `DrawRectangleUsingThreePointsMode` implementation for geo or cartesian mode, building on the updates from this PR#641 - https://github.com/visgl/deck.gl-community/pull/654

PR #643, opened 2026-05-25
fix: double-offset fallback with terrainextension

### LINKED ISSUE #642: [Bug] getMapCoords fallback double-offsets vertices when used with `TerrainExtension`
### Module

- [ ] deck.gl-community/react

### Description

When `EditableGeoJsonLayer` is used together with deck.gl's `_TerrainExtension` and a terrain source layer (e.g. `PointCloudLayer` with `operation: 'terrain+draw'`), dragging or hovering a vertex over a pixel with no terrain data causes the vertex to fly off-screen.

The root cause is an inconsistency in `getMapCoords()` inside `editable-layer.ts`. When a layer with `pickable === '3d'` is present, the method calls `deck.pickObject({ unproject3D: true })` and **correctly strips the Z component** before returning, because `TerrainExtension` will apply the terrain height offset at render time — keeping Z in the returned coordinate would cause double-offsetting. However, when no terrain pick hit is found (e.g. the cursor is over a no-data gap in the point cloud), the method falls back to `viewport.unproject([x, y])`, which returns a **3-component position** including Z. That Z is the orbit target elevation (e.g. ~200 m). This 3D coordinate is stored verbatim in the GeoJSON feature, and `TerrainExtension` then adds the terrain height on top of it — resulting in the vertex rendering at approximately twice the terrain elevation and jumping off-screen.

The fallback path was never updated to apply the same Z-stripping logic that the terrain pick path uses.

### Expected Behavior

**Expected:** the vertex stays at approximately the terrain elevation of the surrounding area.

**Actual:** the vertex immediately flies off-screen (renders at ~2× the terrain elevation). The corrupted coordinate is committed to state, leaving the vertex permanently unreachable.

### Steps to Reproduce

1. Set up a `DeckGL` scene with an `OrbitView` and `coordinateSystem: 'cartesian'`.
2. Add a `PointCloudLayer` with `operation: 'terrain+draw'` and `pickable: '3d'` as the terrain source. Use data that has regions with no point cloud coverage (no-data gaps).
3. Add an `EditableGeoJsonLayer` with `terrainDrawMode: 'offset'` and `extensions: [new TerrainExtension()]`.
4. Draw a polygon, then switch to `ModifyMode` and drag a vertex over a no-data gap in the point cloud.

### Proposed Fix

In `editable-layer.ts`, `getMapCoords()`, strip Z from the fallback path for consistency with the terrain pick path:

```ts
  const pos = this.context.viewport.unproject([screenCoords[0], screenCoords[1]]);
  return [pos[0], pos[1]] as Position;
```

The comment on the terrain pick path ("Return only lon/lat — discard Z so that TerrainExtension can drape geometry onto the surface without double-offsetting") applies equally to the fallback path. Any time a `pickable: '3d'` layer is registered, the presence of `TerrainExtension` must be assumed and Z must be stripped.

### Environment

- Framework version:
- Browser:
- OS:

### Logs

_No response_

marcomuser:
Fixes #642

### DIFF
--- a/modules/editable-layers/src/editable-layers/editable-layer.ts
+++ b/modules/editable-layers/src/editable-layers/editable-layer.ts
@@ -278,6 +278,11 @@ export abstract class EditableLayer<
           // can drape geometry onto the surface without double-offsetting.
           return [pickInfo.coordinate[0], pickInfo.coordinate[1]] as Position;
         }
+        // No terrain hit (e.g. no-data gap) — still strip Z for the same reason:
+        // TerrainExtension is present whenever pickable:'3d' layers exist and will
+        // add terrain height at render time, so pre-baking Z would double-offset.
+        const pos = this.context.viewport.unproject([screenCoords[0], screenCoords[1]]);
+        return [pos[0], pos[1]] as Position;
       }
     }
     return this.context.viewport.unproject([screenCoords[0], screenCoords[1]]) as Position;
--- a/modules/editable-layers/src/editable-layers/editable-layer.ts
+++ b/modules/editable-layers/src/editable-layers/editable-layer.ts
@@ -273,16 +273,13 @@ export abstract class EditableLayer<
           layerIds,
           unproject3D: true
         });
-        if (pickInfo?.coordinate) {
-          // Return only lon/lat — discard Z so that TerrainExtension
-          // can drape geometry onto the surface without double-offsetting.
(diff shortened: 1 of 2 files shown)


charlieforward9 (approved):
Verified locally after tightening the fallback path and adding regression coverage for both terrain-hit and no-hit coordinates.

PR #672, opened 2026-07-01
feat(editable-layers) Snap to edges - alternative approach with Snapper interface

jake-goldsmith:
See https://github.com/visgl/deck.gl-community/pull/623 for more context -- this is an alternative approach which adds in a new Snapper interface to allow developers to use their own custom snapping logic if required.
```
/**
 * Represents the calculation to determine the appropriate snapping point for a given coordinate.
 * Can be overridden to allow custom snapping for different use cases which might have specific requirements around for example optimisation.
 */
export interface Snapper {
  /**
   * Snaps the given event to the nearest snapping point based on the provided candidate features.
   */
  snap(
    event: BasePointerEvent,
    props: ModeProps<FeatureCollection>,
    excludedFeatureIndexes: Set<number>
  ): SnapResult | null;
}
```

We've used a similar interface to snap to our maplibre baselayer. It's helpful to have this in an interface on top of the additionalFeatures property because the maplibre baselayer has a built-in index so we need to be able to query features around the cursor position. Running it in the interface avoids any kind of synchronisation issues if we tried to poke features into the additionFeatures whenever the cursor moves.


charlieforward9:
Confirmed the failing `test-node` job completed the node tests successfully and then failed while trying to write the coverage comment from a fork PR token. Opened #673 to make fork PRs write coverage to the Actions summary and skip the comment write path.

PR #677, opened 2026-07-02
fix(editable-layers): edit mode constructor types (#676)

### LINKED ISSUE #676: [Bug] Editable-layers - TS incompatible types between edit modes
### Module

- [ ] deck.gl-community/react

### Description

We run a project which uses the editable-layers module and we run with strict typing turned on in our typescript config to catch typing errors at compile time.

When we try to upgrade to v9.3 we are seeing type compatibility errors for various edit modes because they aren't correctly extending the function signatures from `GeoJsonEditMode`, they are using different types.

For example with `TranslateMode` we get the error message:

```
 ERROR(TypeScript)  Type 'typeof TranslateMode' is not assignable to type 'GeoJsonEditMode | typeof GeoJsonEditMode | undefined'.
  Type 'typeof TranslateMode' is not assignable to type 'typeof GeoJsonEditMode'.
    Construct signature return types 'TranslateMode' and 'GeoJsonEditMode' are incompatible.
      The types of 'handleStartDragging' are incompatible between these types.
        Type '(event: StartDraggingEvent, props: ModeProps<SimpleFeatureCollection>) => void' is not assignable to type '(_event: StartDraggingEvent, _props: ModeProps<FeatureCollection<Geometry, GeoJsonProperties>>) => void'.
          Types of parameters 'props' and '_props' are incompatible.
            Type 'ModeProps<FeatureCollection<Geometry, GeoJsonProperties>>' is not assignable to type 'ModeProps<SimpleFeatureCollection>'.
              Types of property 'data' are incompatible.
                Type 'FeatureCollection<Geometry, GeoJsonProperties>' is not assignable to type 'SimpleFeatureCollection'.
                  Type 'Geometry' is not assignable to type 'SimpleGeometry'.
                    Type 'GeometryCollection<Geometry>' is not assignable to type 'SimpleGeometry'.
                      Property 'coordinates' is missing in type 'GeometryCollection<Geometry>' but required in type 'MultiPolygon'.
```

This is because the `handleStartDragging` function is defined on `GeoJsonEditMode` as taking `props` of type `ModeProps<FeatureCollection>`  but in `TranslateMode` it's instead of type `ModeProps<SimpleFeatureCollection>` which isn't compatible.

Just using this example for illustrative purposes, this affects lots of other modes and overridden functions too.

Proposed fix - we could either:
* turn on the config `"strict": true` in our tsconfig file. We could either do this for the whole project or just for editable-layers to reduce the effort to fix all the type inconsitencies. Cons - there are other type issues which would need fixing which don't impact consumers because they are internal to the library.
* Audit all classes which extend GeoJsonEditMode and check that they are using compatible types.

I'm also not sure whether it's best to update all of the extenders of `GeoJsonEditMode` or if we can update at the top level - does `GeoJsonEditMode` ever actually need to handle GeometryCollections which seems to be the difference between `FeatureCollection` and `SimpleFeatureCollection`?

### Expected Behavior

I expect that all classes that inherit from `GeoJsonEditMode` extend it with compatible types.

### Steps to Reproduce

The easiest way to see the issues without setting up a whole new project is to enable strict mode in the tsconfig configuration and then run the compilation which should print all of the issues:

```
yarn tsc -p modules/editable-layers/tsconfig.json
```

(although this will also print other internal issues which don't necessarily need to be fixed)

### Environment

Upgrading to @deck.gl-community/editable-layers version 9.3.7

### Logs

_No response_

charlieforward9:
## Summary

Fixes #676.

Aligns the editable-layers public edit-mode type contract with the actual editable geometry surface. `GeoJsonEditModeType`, `GeoJsonEditMode`, `CompositeMode`, `EditableGeoJsonLayer`, and exported edit-mode entry points now use `SimpleFeatureCollection` instead of the wider GeoJSON `FeatureCollection` that includes `GeometryCollection`. This prevents strict TypeScript consumers from seeing exported mode constructors like `TranslateMode` as incompatible with `GeoJsonEditModeConstructor`.

Also adds a focused export/type regression in `imports.node.spec.ts` that assigns the common exported mode constructors to `GeoJsonEditModeConstructor[]`, plus release/upgrade notes for TypeScript consumers in `docs/whats-new.md` and `docs/upgrade-guide.md`.

## Classification

Bug fix: TypeScript declaration/API compatibility for editable-layers mode constructors.

Visual proof is not applicable. This is a type-only compatibility fix with no rendered, interactive, or example behavior change.

## Validation

- `yarn`
- `yarn tsc -p modules/editable-layers/tsconfig.json --noEmit`
- `yarn test-node modules/editable-layers/test/imports.node.spec.ts`
- `yarn test-node modules/editable-layers/test`
- `yarn lint`
- pre-commit `yarn test`: 163 files passed, 1380 tests passed, 9 skipped

Strict-mode reproduction note:

- `yarn tsc -p modules/editable-layers/tsconfig.json --strict --noEmit` still fails on pre-existing internal strictness issues, but no longer reports edit-mode constructor incompatibility or edit-mode `GeoJsonEditMode` override incompatibility. The remaining `TS2416` hits are unrelated widget `deck` nullability overrides.

## Manual verification

No UI route required. Review the generated declarations/build output from `yarn build`, the type assertion in `modules/editable-layers/test/imports.node.spec.ts`, and the v9.4 TypeScript migration note in `docs/upgrade-guide.md`.

## Residual risk

This narrows the public editable data contract away from `GeometryCollection`. That matches the existing `SimpleFeatureCollection` edit-mode implementation and avoids promising support for geometries the edit logic does not handle.

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -6,6 +6,14 @@ Please refer the documentation of each module for detailed upgrade guides.
 
 ## v9.4
 
+### `@deck.gl-community/editable-layers`
+
+- TypeScript change: edit mode constructors now use `SimpleFeatureCollection`
+  consistently across the base classes and exported modes. Consumers that typed
+  editable mode data as the broader GeoJSON `FeatureCollection` should migrate
+  those edit-mode generics to `SimpleFeatureCollection` or `SimpleFeature` from
+  `@deck.gl-community/editable-layers`.
+
 ### `@deck.gl-community/panels`
 
 - Breaking change: panel composition APIs no longer expose `Widget*` names from `@deck.gl-community/panels`.
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -9,6 +9,13 @@ Scope tracked in the [v9.4 milestone](https://github.com/visgl/deck.gl-community
 - `DependencyArrowLayer` - NEW directional marker layer for dependency links with path, line, or arc routing.
 - `PathOutlineLayer` and `PathMarkerLayer` now use deck.gl v9-native sublayers for outlined paths, dashed strokes, and pixel-sized directional markers, restoring the path outline and marker example.
 
+### `@deck.gl-community/editable-layers`
+
+- Edit mode constructor and base-class types now use the package's editable
+  `SimpleFeatureCollection` contract consistently, so exported edit mode
+  constructors can be assigned to `GeoJsonEditModeConstructor[]` in strict
+  TypeScript projects.
+
 ### `@deck.gl-community/infovis-layers`
 
 - Added generic animation, block, fast-text, UTF8 Arrow string-view, view-layout, and viewport-bounds helpers for trace-style visualizations.
--- a/modules/editable-layers/src/edit-modes/composite-mode.ts
+++ b/modules/editable-layers/src/edit-modes/composite-mode.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 // Copyright (c) vis.gl contributors
 
-import {FeatureCollection} from '../utils/geojson-types';
+import {SimpleFeatureCollection} from '../utils/geojson-types';
 import {
   ModeProps,
   ClickEvent,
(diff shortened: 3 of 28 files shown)


charlieforward9:
@jake-goldsmith I wonder if `FeatureCollection` is necessary within this package since `GeometryCollection` isnt really in the scope of what any of these modes can do. If this is the case, we can probably remove the export from utils and simplify the logic further! Let me know thoughts.

jake-goldsmith:
> I wonder if FeatureCollection is necessary within this package since GeometryCollection isnt really in the scope of what any of these modes can do. If this is the case, we can probably remove the export from utils and simplify the logic further! Let me know thoughts.

That sounds like a good idea yeah - I'm not sure if it needs to be done now though or it can be deferred. I'm also tempted to go through and try to clean up some of the other type inconsistencies but likewise that is probably worth waiting for another ticket.

My only concern with these changes is whether it's going to put a burden on developers using typescript to fix their types - I imagine most would already be using the types from geojson so I guess it's hopefully just a case of swapping `Feature` with `SimpleFeature` and `FeatureCollection` with `SimpleFeatureCollection` - probably worth at least calling out in the release notes etc. so users are aware?

charlieforward9:
> burden on developers using typescript to fix their types

good call... Ill wait on the [deck9.4 release](https://github.com/visgl/deck.gl/pull/10409) to follow suite and push this through while adding the relevant docs

charlieforward9:
Updated after the deck.gl 9.4 docs PR landed: rebased this branch onto current master and added v9.4 release/upgrade notes for the editable-layers TypeScript contract. Local validation passed with `yarn`, editable-layers type/test gates, `yarn lint`, `cd website && yarn && yarn build`, and pre-commit `yarn test`.

charlieforward9 (commented):
This can be merged across the v9.4 release.

I'm curious if `FeatureCollection` is necessary at all to export anymore? Could we set a deprecation on it and remove it entirely by v10?

PR #679, opened 2026-07-13
fix(editable-layers): use named turf imports to fix .default is not a function in CJS/esbuild consumers

igorDykhta:
## Summary

Switch every `@turf/*` import in editable-layers from a default import to the
equivalent named import (`import turfX from '@turf/x'` → `import {x as turfX} from '@turf/x'`).
Local names and call sites are unchanged.

turf v7 exposes each function as a named export; its `default` only resolves when
a bundler's ESM/CJS interop preserves it. When editable-layers is consumed through
a CommonJS package and re-bundled by esbuild, turf gets wrapped as
`__toESM(require("@turf/x"), 1)`, making `.default` a namespace object rather than
the function, so `(0, import_x.default)(...)` throws `.default is not a function`
at runtime (breaks rectangle/polygon drawing, rewind, etc.).

Named imports don't depend on interop heuristics and match how the codebase already
imports `@turf/helpers`.

## Note on reproduction

This does not reproduce via the Vite-based examples here (Rollup's interop preserves
`.default`). It reproduces when editable-layers is bundled with esbuild through a CJS
consumer — e.g. `new DrawRectangleMode().getTwoClickPolygon(...)` throws
`(0, import_bbox_polygon.default) is not a function`. The named imports run cleanly
in that same pipeline.

### DIFF
--- a/modules/editable-layers/src/edit-modes/coordinate-system.ts
+++ b/modules/editable-layers/src/edit-modes/coordinate-system.ts
@@ -4,8 +4,8 @@
 
 import {bearing as turfBearing} from '@turf/bearing';
 import {distance as turfDistance} from '@turf/distance';
-import {destination as turfDestination} from '@turf/destination';
-import {midpoint as turfMidpoint} from '@turf/midpoint';
+import {destination} from '@turf/destination';
+import {midpoint} from '@turf/midpoint';
 import {point} from '@turf/helpers';
 import {COORDINATE_SYSTEM} from '@deck.gl/core';
 import type {CoordinateSystem} from '@deck.gl/core';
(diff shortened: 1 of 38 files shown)


charlieforward9 (approved):
Pretty straightforward.  

Nitpicking: some files import the method names regularly, while others have the turf* suffix. 

What's the impact of normalizing everything to just the name of the method. Eg. not casting?

igorDykhta:
> Nitpicking: some files import the method names regularly, while others have the turf* suffix.
> 
> What's the impact of normalizing everything to just the name of the method. Eg. not casting?

I wanted to touch only imports, but I'm happy to normalize the non-colliding ones if we prefer some consistency here.

PR #716, opened 2026-08-16
feat(three): render detailed palms

charlieforward9:
## Summary

- replace the generic flattened palm crown with a deterministic feathered palm silhouette
- add a tapered, ring-scarred trunk whose shaft terminates inside the crown heart
- document the improved palm geometry and lock it down with focused regression tests
- cover crown density, geometry bounds, determinism, and the trunk/crown connection

## Why

`TreeLayer`'s existing `palm` geometry read as a flattened canopy rather than a recognizable
palm. It also left the trunk and crown as independently scaled meshes, which could expose a
visible gap for some trunk-to-canopy ratios.

## Implementation

The shared procedural crown now contains 20 mature arching fronds, 8 upright spear fronds, and
284 pairs of tapered pinnate leaflets. It is generated once and remains one instanced canopy
mesh. Palms use a dedicated ring-scarred trunk mesh; mixed forests pay only one additional
instanced draw call, while non-palm tree geometry is unchanged.

The palm shaft extends by 24% of the canopy height, ending inside the crown heart. Because that
overlap is canopy-relative, the connection stays closed as tree proportions vary.

## Visual verification

| Before | After |
|---|---|
| Before | After |

- inspected the Palm Grove in the live Wild Forest example
- checked varied palm heights and trunk fractions at close range; no crown/trunk daylight remains
- confirmed a clean browser console

Review locally with `yarn start-local` from `examples/three/wild-forest`, then open
`http://127.0.0.1:8080/`.

## Validation

- `yarn`
- `yarn build`
- `yarn test-node modules/three/test/tree-layer.spec.ts` — 168 files passed; 1,426 tests passed, 9 skipped
- `yarn lint-fix`
- `yarn lint`
- `git diff --check`

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -103,10 +103,7 @@ Scope tracked in the [v9.4 milestone](https://github.com/visgl/deck.gl-community
 
 ### `@deck.gl-community/three`
 
-- `TreeLayer` now renders its `palm` silhouette as a recognizable date palm with a tapered,
-  ring-scarred trunk, a connected crown shaft, 20 mature arching fronds, 8 upright spear fronds,
-  and paired pinnate leaflets. The procedural crown remains one shared instanced mesh, and the
-  Wild Forest example includes a dedicated Date Palm Grove for visual inspection.
+- `TreeLayer`: improved `palm` silhouette with a detailed frond crown and ring-scarred trunk.
 
 ### `@deck.gl-community/widgets`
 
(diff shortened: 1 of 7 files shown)


docs/whats-new.md
charlieforward9, line 1: @copilot no need to be so verbose...

charlieforward9 (commented):
@copilot - still looks like a palm - so lets not be so gramatically restrictive to dates. Keep the docs concise and focus on clean, efficient code and a minimal changelog here.

PR #718, opened 2026-08-28
WIP revert

ibgreen:
Revert trace-layers.


charlieforward9:
Why the revert? Is this not stable enough to use. a competing lib already solve this? A team member is building a trace pipeline now I was curious to utilize this for so some rationale for removing would be helpful. @ibgreen

PR #732, opened 2026-09-04
fix(editable-layers): remove layers peerDependency

felixpalmer:
Nothing in `@deck.gl-community/editable-layers` references `@deck.gl-community/layers` (since the deletions in https://github.com/visgl/deck.gl-community/pull/447). By listing as a peer, consumers of `@deck.gl-community/editable-layers` are incorrectly prompted to include  `@deck.gl-community/layers` which they do not need

### Changes

- Remove `@deck.gl-community/layers` peerDependency

### DIFF
--- a/modules/editable-layers/package.json
+++ b/modules/editable-layers/package.json
@@ -77,7 +77,6 @@
     "preact": "^10.17.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


charlieforward9:
Good catch!

Issue #747, opened 2026-09-15
Tracker: react-fiber module


ibgreen:
## Purpose

Track the follow-up work for `@deck.gl-community/react-fiber` after the direct source fork in [PR #746](https://github.com/visgl/deck.gl-community/pull/746).

This package is incubating in deck.gl-community as a deck.gl 9.4 addon. The long-term direction is for it to replace `@deck.gl/react` in deck.gl v10. This tracker covers the convention-alignment work needed before that migration; it does not change the direct-fork/no-wrapper principle.

## Findings

The current fork preserves Brandon Pierce's `fiber.gl` 2.0.1 implementation and its reconciler-oriented API, but its public React surface differs from deck.gl/react:

- The canonical component should be `DeckGL`, with `DeckGLProps` and `DeckGLRef`; retain `Deckgl`, `DeckglProps`, and `useDeckgl` as deprecated incubation aliases.
- The ref should match the official React contract: `deck` plus `pickObjectAsync`, `pickObjectsAsync`, `pickObject`, `pickObjects`, and `pickMultipleObjects`.
- The component should use the official prop filtering and sizing conventions, including React-owned `width`, `height`, `style`, `id`, and `className`, while preserving fiber-specific `interleaved` support.
- Add the official context-provider contract for viewport, container, event manager, view-state changes, deck, and widgets.
- Support ordinary React children, view-scoped child positioning, render callbacks, and JSX layers/views/widgets.
- Support conventional PascalCase JSX such as `<ScatterplotLayer />` and `<MapView />`, while retaining universal lowercase `<layer />` and `<view />`, legacy typed JSX, `extend`, and the direct reconciler entrypoints.
- Make the package root follow deck.gl/react conventions with default and named `DeckGL` exports plus the convention-aligned public types.
- Update the module docs and all copied examples to teach the canonical API and provide a migration path from the incubator names.

The official reference behavior is documented in [DeckGL React Component](https://deck.gl/docs/api-reference/react/deckgl) and implemented in [deck.gl's React source](https://github.com/visgl/deck.gl/tree/master/modules/react/src).

## Follow-up PR acceptance criteria

- No runtime dependency on `@deck.gl/react` or `@deckgl-fiber-renderer/*`; the module remains a direct source fork with local implementation.
- Existing upstream tests continue to pass without reduced coverage.
[...]

PR #761, opened 2026-09-18
react-fiber -> @deck.gl/react backwards compatibility 

brandonjpierce:
This is a carry over from https://github.com/deckgl-fiber-renderer/fiber.gl/pull/20

ℹ️ AI was used to create this code. Original spec documents were produced via `openspec`.

---

## Add a bounded `@deck.gl/react` compatibility API

This PR adds an additive compatibility API to `@deck.gl-community/react-fiber`. It helps supported `@deck.gl/react` applications migrate without replacing the existing native React Fiber API.

Applications can now replace:

```ts
import {DeckGL} from '@deck.gl/react';
```

with:

```ts
import {DeckGL} from '@deck.gl-community/react-fiber/compat';
```

The adapter lowers supported React-style layer and view components to the existing native `<layer>` and `<view>` reconciler primitives. The root package API remains unchanged: `Deckgl`, `useDeckgl`, and `extend` continue to provide the native renderer interface.

## What changed

The package now publishes these additive subpaths:

- `@deck.gl-community/react-fiber/compat`
- `@deck.gl-community/react-fiber/compat/layers`
- `@deck.gl-community/react-fiber/compat/geo-layers`
- `@deck.gl-community/react-fiber/compat/aggregation-layers`
- `@deck.gl-community/react-fiber/compat/mesh-layers`

The root compat entry point exports `DeckGL`, five view wrappers, and the bounded `DeckGLProps`, `DeckGLRef`, and `DeckGLContextValue` types. The layer subpaths export only the approved wrapper matrices. The geo-layer entry point also provides the legacy `mVTLayer`/`mvtLayer` and `wMSLayer`/`wmsLayer` aliases.

`DeckGL` delegates renderer setup to the existing `Deckgl` component. It adds a root-specific ref and context bridge so each compat root exposes its own `Deck` or `MapboxOverlay` instance instead of relying on the legacy shared `useDeckgl()` store.

The ref surface is limited to:

- `deck`
- `pickObject`
- `pickObjects`
- `pickMultipleObjects`
- `pickObjectAsync`
- `pickObjectsAsync`

Before initialization, `deck` is `null` and picking methods throw. A supplied `ContextProvider` receives only `{deck}`. The adapter does not claim support for the broader `@deck.gl/react` context contract.

In development, compat warns when it detects unsupported renderer ownership props (`gl`, `canvas`, `parent`, or `_customRender`) or function children. Production remains silent.

## Scope boundaries

This is a migration aid, not full `@deck.gl/react` parity.

The compat API does not add widgets, `useWidget`, widget wrappers, unrestricted imperative APIs, unrestricted context APIs, or wrapper components outside the approved layer and view matrices. It also retains the existing interleaved `MapboxOverlay` view limitation. Compat view wrappers do not change that behavior.

Applications that need low-level canvas, WebGL, or parent ownership should continue to use the native `Deckgl` component.

## Documentation and tests

The PR adds a migration guide, package README guidance, docs sidebar wiring, API/get-started cross-links, and a release-note entry. The guide documents the supported export matrix, stable layer and view IDs, ref and context behavior, unsupported patterns, and interleaved-rendering limits.

The test suite covers:

- root and subpath export matrices;
- approved aliases and excluded aggregation internals;
- layer and view lowering to native reconciler primitives;
- ref method delegation;
- bounded context delivery;
- development warnings and production silence; and
- compat prop exclusions.

## Validation

- `yarn workspace @deck.gl-community/react-fiber test`
- `yarn tsc --project modules/react-fiber/tsconfig.json --noEmit`
- `yarn build`
- `yarn lint`
- `yarn --cwd website build`
- Verified every compat export resolves in ESM, CommonJS, and declaration form.

The website build completed with existing non-blocking warnings from unrelated
examples and documentation.

<details>
<summary>Semantic diff</summary>

```md
### docs/modules/react-fiber/README.md

| Status | Type | Name |
|--------|------|------|
| Δ | heading | @deck.gl-community/react-fiber |

### docs/modules/react-fiber/api-reference/deckgl.md

| Status | Type | Name |
|--------|------|------|
| Δ | heading | Deckgl |
| + | heading | Migrating from `@deck.gl/react` |

### docs/modules/react-fiber/developer-guide/get-started.md

| Status | Type | Name |
|--------|------|------|
| + | heading | Migrating from `@deck.gl/react` |

### docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md

| Status | Type | Name |
|--------|------|------|
| + | heading | Migrate from `@deck.gl/react` |
| + | heading | Replace the root import |
| + | heading | Supported exports |
| + | heading | Render layers and views |
| + | heading | Ref and context |
| + | heading | Review unsupported patterns |

### docs/modules/react-fiber/sidebar.json

| Status | Type | Name |
|--------|------|------|
| Δ | array | items |

### docs/whats-new.md

| Status | Type | Name |
|--------|------|------|
| Δ | heading | `@deck.gl-community/react-fiber` |

### modules/react-fiber/README.md

| Status | Type | Name |
|--------|------|------|
| + | heading | Migrate from `@deck.gl/react` |

### modules/react-fiber/package.json

| Status | Type | Name |
|--------|------|------|
| + | property | types |
| + | property | require |
| + | property | import |
| + | property | types |
| + | property | require |
| + | property | import |
| + | property | types |
| + | property | require |
| + | property | import |
| + | property | types |
| + | property | require |
| + | property | import |
| + | property | types |
| + | property | require |
| + | property | import |
| + | property | @deck.gl/aggregation-layers |
| + | property | @deck.gl/aggregation-layers |

### modules/react-fiber/src/compat/__tests__/deckgl.test.tsx

| Status | Type | Name |
|--------|------|------|
| + | variable | deck |
| + | property | children |
| + | property | onDeckglChange |
| + | variable | {DeckGL} |
| + | test_suite | DeckGL compatibility adapter |
| + | test_hook | beforeEach |
| + | test_hook | afterEach |
| + | test | exposes the bounded ref surface and delegates picking after initialization |
| + | test | provides only the supported deck context value |
| + | test | warns only in development for unsupported renderer props and function children |

### modules/react-fiber/src/compat/__tests__/entrypoints.test.ts

| Status | Type | Name |
|--------|------|------|
| + | test_suite | compat entrypoint export matrices |
| + | test | exports only the approved root compatibility API |
| + | test | exports the approved layer family wrappers and no aggregation internals |
| + | test | exports the approved geo and mesh wrappers, including geo aliases |

### modules/react-fiber/src/compat/__tests__/types.test-d.ts

| Status | Type | Name |
|--------|------|------|
| + | variable | props |
| + | property | deck |
| + | variable | unsupportedCanvas |

### modules/react-fiber/src/compat/__tests__/wrappers.test.tsx

| Status | Type | Name |
|--------|------|------|
| + | test_suite | compat wrappers |
| + | test | lowers a layer component to the native layer primitive |
| + | test | lowers a view component and its nested children to the native view primitive |

### modules/react-fiber/src/compat/aggregation-layers.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | LayerComponent |
| + | type | LayerConstructor |
| + | variable | ScreenGridLayer |
| + | variable | HexagonLayer |
| + | variable | ContourLayer |
| + | variable | GridLayer |
| + | variable | HeatmapLayer |

### modules/react-fiber/src/compat/context.ts

| Status | Type | Name |
|--------|------|------|
| + | variable | EMPTY_CONTEXT_VALUE |

### modules/react-fiber/src/compat/deckgl.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | UnsupportedProps |
| + | type | RuntimeDeckGLProps |
| + | type | PickingMethod |
| + | function | getDeckOrThrow |
| + | function | callPickingMethod |
| + | function | warnForUnsupportedUsage |
| + | variable | DeckGL |

### modules/react-fiber/src/compat/geo-layers.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | LayerComponent |
| + | type | LayerConstructor |
| + | variable | S2Layer |
| + | variable | QuadkeyLayer |
| + | variable | TileLayer |
| + | variable | H3ClusterLayer |
| + | variable | H3HexagonLayer |
| + | variable | Tile3DLayer |
| + | variable | TerrainLayer |
| + | variable | GeohashLayer |
| + | variable | GreatCircleLayer |
| + | variable | TripsLayer |
| + | variable | MVTLayer |
| + | variable | WMSLayer |
| + | variable | mVTLayer |
| + | variable | mvtLayer |
| + | variable | wMSLayer |
| + | variable | wmsLayer |

### modules/react-fiber/src/compat/index.ts

| Status | Type | Name |
|--------|------|------|
| + | export | DeckGL |
| + | export | DeckGLContextValue |
| + | export | DeckGLProps |
| + | export | DeckGLRef |
| + | export | FirstPersonView |
| + | export | GlobeView |
| + | export | MapView |
| + | export | OrbitView |
| + | export | OrthographicView |

### modules/react-fiber/src/compat/layer-wrapper.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | LayerConstructor |
| + | function | createLayerComponent |
| + | function | LayerComponent |

### modules/react-fiber/src/compat/layers.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | LayerComponent |
| + | type | LayerConstructor |
| + | variable | ArcLayer |
| + | variable | BitmapLayer |
| + | variable | IconLayer |
| + | variable | LineLayer |
| + | variable | PointCloudLayer |
| + | variable | ScatterplotLayer |
| + | variable | ColumnLayer |
| + | variable | GridCellLayer |
| + | variable | PathLayer |
| + | variable | PolygonLayer |
| + | variable | GeoJsonLayer |
| + | variable | TextLayer |
| + | variable | SolidPolygonLayer |

### modules/react-fiber/src/compat/mesh-layers.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | LayerComponent |
| + | type | LayerConstructor |
| + | variable | ScenegraphLayer |
| + | variable | SimpleMeshLayer |

### modules/react-fiber/src/compat/types.ts

| Status | Type | Name |
|--------|------|------|
| + | type | DeckGLInstance |
| + | property | deck |
| + | type | DeckGLProps |
| + | property | pickObject |
| + | property | pickObjects |
| + | property | pickMultipleObjects |
| + | property | pickObjectAsync |
| + | property | pickObjectsAsync |

### modules/react-fiber/src/compat/views.tsx

| Status | Type | Name |
|--------|------|------|
| + | type | ViewConstructor |
| + | type | ViewComponent |
| + | function | createViewComponent |
| + | function | ViewComponent |
| + | property | children |
| + | property | view |
| + | variable | MapView |
| + | variable | OrthographicView |
| + | variable | OrbitView |
| + | variable | FirstPersonView |
| + | variable | GlobeView |

### modules/react-fiber/src/dom/components.tsx

| Status | Type | Name |
|--------|------|------|
| + | property | onDeckglChange |
| Δ | function | DeckglComponent |
| Δ | function | Deckgl |

**Summary:** 153 added, 6 modified, 1 deleted across 23 files.
```

</details>


brandonjpierce:
Original plan prompt for posterity sake:

```markdown
# React DeckGL Compatibility Adaptation Plan

## 1. Purpose

Adapt the upstream `deckgl-react-compat` OpenSpec change into the existing
`@deck.gl-community/react-fiber` package. The result is an **additive**
compatibility API that lets a supported subset of `@deck.gl/react` applications
replace:

\```ts
import {DeckGL} from '@deck.gl/react';
\```

with:

\```ts
import {DeckGL} from '@deck.gl-community/react-fiber/compat';
\```

The adapter will lower approved React-style layer and view components to this
package's existing native `<layer>` and `<view>` reconciler primitives. It is a
migration aid, not a claim of full `@deck.gl/react` parity. In particular, it
will support the approved wrapper matrices, the six-method ref surface, and
`deck`-only context parity; it will not add widgets, `useWidget`, official
function-child render callbacks, or unrestricted imperative/context APIs.

This plan follows the accepted, narrowed direction in the upstream
`proposal.md`, `design.md`, `tasks.md`, and the post-conflict-marker versions
of the compatibility specs. The upstream change contains unresolved conflicts;
the contradictory broad/full-parity and widget requirements are not adopted.
See [Scope Boundaries and Risks](#5-scope-boundaries-and-risks).

## 2. Source-to-Target Structural Mapping

### Repository differences

Upstream fiber.gl separates the renderer into `packages/dom`,
`packages/reconciler`, `packages/shared`, and `packages/types`, and proposes a
new published `@deckgl-fiber-renderer/compat` package. This repository has
already consolidated the implementation into one workspace package:

\```text
modules/react-fiber/
  src/dom/
  src/reconciler/
  src/shared/
  src/types/
\```

The target must therefore add a **subpath of the existing package**, not create
another workspace or recreate upstream package boundaries. The public package
name is `@deck.gl-community/react-fiber`, and its package map in
`modules/react-fiber/package.json` is the source of truth for every new compat
entrypoint.

### Source-to-target map

| Upstream assumption or location | Target equivalent | Adaptation decision |
| --- | --- | --- |
| New package `@deckgl-fiber-renderer/compat` | `@deck.gl-community/react-fiber/compat` | Add a subpath export to the consolidated package; do **not** add a `modules/compat` workspace. |
| `@deckgl-fiber-renderer/compat/layers`, `/geo-layers`, `/aggregation-layers`, `/mesh-layers` | `@deck.gl-community/react-fiber/compat/layers`, `/compat/geo-layers`, `/compat/aggregation-layers`, `/compat/mesh-layers` | Add four explicit package-map entries and source entry files under `modules/react-fiber/src/compat/`. These are migration-convenience APIs, not attempted subpaths of `@deck.gl/react`. |
| `packages/dom/src/components.tsx` and `hooks.tsx` | `modules/react-fiber/src/dom/components.tsx` and `hooks.tsx` | Reuse the existing `Deckgl` renderer component and `useDeckgl()` bridge; do not duplicate Deck/MapboxOverlay setup. |
| `packages/reconciler/src/config.ts`, `renderer.ts`, and `utils.ts` | `modules/react-fiber/src/reconciler/config.ts`, `renderer.ts`, and `utils.ts` | Reuse the existing native `<layer>`/`<view>` construction, tree flattening, `Deck`/`MapboxOverlay` selection, and interleaved-view constraint. Do not modify this stable reconciler unless a compat integration test demonstrates a real gap. |
| `packages/shared/src/store.ts` | `modules/react-fiber/src/shared/store.ts` | Existing `deckgl` state is the only available bridge for the compat ref/context. It does not provide the full official `@deck.gl/react` context. |
| `packages/types/src/react.ts` and `jsx.ts` | `modules/react-fiber/src/types/react.ts` and `jsx.ts` | Keep native `DeckglProps` and intrinsic-element types intact. Define bounded compat-specific public types under the new `src/compat/` subtree; do not expose legacy intrinsic aliases as compat wrapper APIs. |
| Upstream package-local tests | `modules/react-fiber/src/compat/__tests__/` plus existing `src/dom`, `src/reconciler`, and `src/types` tests | Add compat-focused tests beside the new source. Reuse existing renderer/reconciler fixtures instead of creating a second test harness. |
| Upstream migration/package documentation | Package README plus `docs/modules/react-fiber/` | Document the `@deck.gl-community/react-fiber/compat` import and subpaths. Keep the existing site sidebar convention. |
| Upstream changeset requirement | Repository release process in `AGENTS.md` and `docs/whats-new.md` | This repository does not use Changesets. Record the public API in `docs/whats-new.md`; version/tag through the existing `yarn publish-beta` or `yarn publish-prod` process when the change is released. |

### Upstream assumptions that do not apply

- There will be no standalone `@deckgl-fiber-renderer/compat` package and no
  restored four-package architecture.
- The upstream `design.md` reference to `@deckgl-fiber-renderer/dom/compat`
  is contradictory and does not define a target import. The only target import
  contract is `@deck.gl-community/react-fiber/compat`.
- No new example application is required. Upstream explicitly treats adding
  example apps as a non-goal; existing `examples/react-fiber/*` applications
  remain the manual consumer environments.
- The target must not infer that official `@deck.gl/react` has corresponding
  `/layers` or `/geo-layers` subpaths. The compat subpaths are intentionally
  new APIs owned by this package.

## 3. Implementation Plan

### 1. Lock the translated public contract before implementation

**Targets:** `PLAN.md` (this plan), then the implementation PR description and
public API review; no runtime source change in this step.

**Change:** Adopt the following bounded API as the acceptance contract:

- root: `DeckGL`, `MapView`, `OrthographicView`, `OrbitView`,
  `FirstPersonView`, `GlobeView`, and the named types `DeckGLProps`,
  `DeckGLRef`, and `DeckGLContextValue`;
- layer-wrapper subpaths and exact exports:
  - `compat/layers`: `ArcLayer`, `BitmapLayer`, `IconLayer`, `LineLayer`,
    `PointCloudLayer`, `ScatterplotLayer`, `ColumnLayer`, `GridCellLayer`,
    `PathLayer`, `PolygonLayer`, `GeoJsonLayer`, `TextLayer`,
    `SolidPolygonLayer`;
  - `compat/geo-layers`: `S2Layer`, `QuadkeyLayer`, `TileLayer`,
    `H3ClusterLayer`, `H3HexagonLayer`, `Tile3DLayer`, `TerrainLayer`,
    `GeohashLayer`, `GreatCircleLayer`, `TripsLayer`, `MVTLayer`, `WMSLayer`;
  - `compat/aggregation-layers`: `ScreenGridLayer`, `HexagonLayer`,
    `ContourLayer`, `GridLayer`, `HeatmapLayer`;
  - `compat/mesh-layers`: `ScenegraphLayer`, `SimpleMeshLayer`.
- aliases from the native type surface: `mVTLayer` and `mvtLayer` map to
  `MVTLayer`; `wMSLayer` and `wmsLayer` map to `WMSLayer`.
- `DeckGLRef` exposes exactly `deck`, `pickObject`, `pickObjects`,
  `pickMultipleObjects`, `pickObjectAsync`, and `pickObjectsAsync`.

**Upstream requirement:** `proposal.md` API surface and migration scope;
`design.md` Decisions 3, 7, 8, and 9; `tasks.md` §1; and the accepted portions
of all six compatibility specs.

**Compatibility consideration:** This review resolves the upstream conflict
markers in favor of the explicitly bounded proposal/design. Do not add exports
for `_AggregationLayer`, `WebGLAggregator`, `CPUAggregator`, aggregation helper
types, widgets, or `useWidget`.

### 2. Add a consolidated compat source boundary and explicit package exports

**Targets:**

- new `modules/react-fiber/src/compat/index.ts`;
- new `modules/react-fiber/src/compat/layers.tsx`;
- new `modules/react-fiber/src/compat/geo-layers.tsx`;
- new `modules/react-fiber/src/compat/aggregation-layers.tsx`;
- new `modules/react-fiber/src/compat/mesh-layers.tsx`;
- `modules/react-fiber/package.json`.

**Change:** Establish `src/compat/` as an internal source grouping within the
existing workspace package. Its `index.ts` is the root compat entrypoint; the
four family files are the only wrapper-family entrypoints. Add the five matching
conditional `exports` entries, each with `types`, `import`, and `require`
targets under `dist/compat/` using the package's existing export-map pattern.
Do not change the existing root, `/dom`, `/reconciler`, `/shared`, or `/types`
exports.

**Upstream requirement:** `proposal.md` “API Surface”; `design.md` Decision 4;
`tasks.md` §2.1 and §3.1; `compat-jsx-layers/spec.md` wrapper-family scenarios.

**Compatibility/export consideration:** Adding subpaths is additive, but every
source entry must build to its declared `dist/compat/*.{js,cjs,d.ts}` artifact.
Do not introduce a new workspace manifest, duplicate peer dependencies, or
alter the existing root entrypoint's default `Deckgl` API.

### 3. Define the bounded compat types and internal context contract

**Targets:**

- new `modules/react-fiber/src/compat/types.ts`;
- new `modules/react-fiber/src/compat/context.ts`;
- `modules/react-fiber/src/compat/index.ts`.

**Change:** Define and root-export `DeckGLProps`, `DeckGLRef`, and
`DeckGLContextValue` without importing types from `@deck.gl/react`.

- Base `DeckGLProps` on the local renderer's applicable deck/map props, but
  intentionally omit `gl`, `canvas`, `parent`, and `_customRender`.
- Add the supported `ContextProvider` hook-up with a type equivalent to a
  provider for `DeckGLContextValue`. The internal context value guarantees only
  `{deck}` (with the appropriate pre-initialization optional/null state).
- Define `DeckGLRef` with the exact six-member inventory from step 1. Type the
  five picking methods as the corresponding local deck instance methods, so
  their sync/async argument and return signatures follow the supported deck.gl
  dependency range.
- Keep any React context object private unless a separate API decision approves
  exporting it. The upstream accepted contract names `DeckGLContextValue`, not
  a root `DeckGLContext` value.

**Upstream requirement:** `design.md` Decisions 7 and 8; `compat-types/spec.md`;
`compat-context/spec.md`; and `compat-ref-methods/spec.md`.

**Compatibility consideration:** Do not type or document official context
fields that this renderer does not supply (`viewport`, `container`,
`eventManager`, `onViewStateChange`, or `widgets`). This avoids a source-level
or declaration-level dependency on `@deck.gl/react` and prevents accidental
full-parity claims.

### 4. Implement the `DeckGL` adapter, ref bridge, and bounded context provider

**Targets:**

- new `modules/react-fiber/src/compat/deckgl.tsx`;
- `modules/react-fiber/src/compat/index.ts`;
- existing `modules/react-fiber/src/dom/components.tsx` and
  `modules/react-fiber/src/dom/hooks.tsx` only if an integration test proves a
  small exported/internal hook is required.

**Change:** Implement a `forwardRef`-based `DeckGL` adapter that renders the
existing local `Deckgl` component rather than reimplementing renderer setup.
Inside its rendered subtree, bridge the local `useDeckgl()` result into:

1. an imperative ref whose `deck` property reflects the underlying `Deck` or
   `MapboxOverlay`, and whose five methods delegate to that instance's picking
   methods; and
2. the optional caller-supplied `ContextProvider`, with a compat context value
   limited to `deck`.

Treat the low-level excluded props and function children as unsupported
patterns. In development, issue concise migration-oriented warnings when they
are detectably supplied; production must remain silent. The warning text and
documentation must not imply that the unsupported pattern continues to work.
Pass the supported props and children through to local `Deckgl` without changing
its standard or interleaved initialization behavior.

**Upstream requirement:** `compat-deckgl-component/spec.md` root-import,
supported-child, exclusions, development-warning, and production-silence
scenarios; `compat-ref-methods/spec.md`; `compat-context/spec.md`; `tasks.md`
§2.2–2.4 and §5.2.

**Compatibility consideration:** `Deckgl` currently creates an isolated
reconciler store per root while `useDeckgl()` exposes a legacy shared hook. The
adapter must be tested with the actual local lifecycle and must not accidentally
make a second root's ref/context point at the first root. It must also preserve
the current `MapboxOverlay` interleaved behavior rather than attempting to
supply `views` to that overlay.

### 5. Implement shared wrapper mechanics and all approved layer families

**Targets:**

- new `modules/react-fiber/src/compat/layer-wrapper.tsx` (private shared
  factory/helper);
- new `modules/react-fiber/src/compat/layers.tsx`;
- new `modules/react-fiber/src/compat/geo-layers.tsx`;
- new `modules/react-fiber/src/compat/aggregation-layers.tsx`;
- new `modules/react-fiber/src/compat/mesh-layers.tsx`.

**Change:** Build a shared wrapper mechanism that converts each approved
uppercase React component to the local native primitive:

\```tsx
<layer layer={new LayerConstructor(layerProps)} />
\```

Each public family file imports only the constructors in its approved matrix and
exports only the matching wrapper components. Export the four legacy-case
aliases from `geo-layers.tsx` as aliases of the same `MVTLayer`/`WMSLayer`
wrappers, not as separate constructors. Keep component construction/imports
family-local so consumers can import only the needed compat family.

**Upstream requirement:** `design.md` Decision 3 and compatibility matrix;
`compat-jsx-layers/spec.md`; `tasks.md` §3.2–3.4.

**Compatibility consideration:** Wrapper props must not be advertised as a
blanket replacement for all deck.gl classes. Preserve required, stable explicit
layer IDs in all test and documentation examples. Do not expose the native
`<layer>` primitive as a changed API, and do not add aggregation internals or
helper/type re-exports.

### 6. Implement root-only compat view wrappers

**Targets:**

- new `modules/react-fiber/src/compat/views.tsx`;
- `modules/react-fiber/src/compat/index.ts`.

**Change:** Implement and root-export only `MapView`, `OrthographicView`,
`OrbitView`, `FirstPersonView`, and `GlobeView`. Each wrapper constructs its
matching deck.gl view from the supplied view props and lowers it to:

\```tsx
<view view={new ViewConstructor({id, ...viewProps})}>{children}</view>
\```

so approved compat layers can nest beneath a compat view. Keep view wrappers
out of the layer-family subpaths.

**Upstream requirement:** `compat-jsx-views/spec.md`; `proposal.md` root
exports; `design.md` Decision 9; `tasks.md` §4.

**Compatibility consideration:** Require explicit IDs in examples and validate
composition using the existing reconciler's view organization. Do not claim
support for views outside the five-item matrix, and do not relax the existing
interleaved `MapboxOverlay` view limitation.

### 7. Add compat-focused runtime, export, and type coverage

**Targets:**

- new `modules/react-fiber/src/compat/__tests__/deckgl.test.tsx`;
- new `modules/react-fiber/src/compat/__tests__/wrappers.test.tsx`;
- new `modules/react-fiber/src/compat/__tests__/entrypoints.test.ts`;
- new `modules/react-fiber/src/compat/__tests__/types.test-d.ts`;
- selectively extend existing tests only where coverage belongs to the existing
  behavior:
  - `modules/react-fiber/src/dom/__tests__/deckgl-component.test.tsx`;
  - `modules/react-fiber/src/reconciler/__tests__/config-instance-creation.test.ts`;
  - `modules/react-fiber/src/reconciler/__tests__/config-tree-operations.test.ts`;
  - `modules/react-fiber/src/reconciler/__tests__/renderer.test.ts`.

**Change:** Add tests that cover the public contract rather than duplicating
current reconciler tests:

- assert root and subpath export matrices exactly, including the four aliases
  and the absence of excluded aggregation/internal APIs;
- mount `DeckGL` with supported props, stable-ID layers, and explicit-ID views;
  verify the adapter delegates to the existing local renderer and compatible
  layer/view trees reach it;
- verify all six ref members, including delegation to sync and async picking
  methods once a deck is available;
- verify a supplied `ContextProvider` receives only the supported `deck`
  context contract and does not imply unimplemented official fields;
- verify development warnings for detectable unsupported props/function-child
  use, and no warning in production;
- verify nested compat layers beneath each supported view and the existing
  interleaved caveat;
- type-check all approved wrapper imports/matrices, compat props exclusions,
  aliases, ref methods, and the absence of a required `@deck.gl/react`
  dependency.

Use the existing Vitest/jsdom setup in `modules/react-fiber/vite.config.js` and
existing DOM/reconciler mocks or `test-renderer.ts` fixtures as appropriate.
Do not add a new test runner or a browser-only harness solely for compat.

**Upstream requirement:** all scenario specs; `tasks.md` §2.4, §3.4, §4.3,
§5.3, and §6.2.

**Compatibility consideration:** Test the built package exports after build as
well as source-level entrypoint tests. A source import passing while the
conditional package map omits a declaration, CJS, or ESM artifact is a release
blocker.

### 8. Document the supported migration and release-facing API

**Targets:**

- `modules/react-fiber/README.md`;
- `docs/modules/react-fiber/README.md`;
- `docs/modules/react-fiber/developer-guide/get-started.md`;
- new `docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md`;
- `docs/modules/react-fiber/api-reference/deckgl.md`;
- `docs/modules/react-fiber/sidebar.json`;
- `docs/whats-new.md`.

**Change:** Write one focused migration guide and link it from the package and
site documentation. It must include:

- the root import replacement and all four compat family subpath mappings;
- a compatibility/export matrix matching steps 1, 5, and 6;
- stable layer-ID and explicit view-ID examples using the target package name;
- the exact six-member ref inventory and a `deck`-only context description;
- an exclusions-review checklist for `gl`, `canvas`, `parent`, and
  `_customRender`;
- clear limitations and workarounds: function-child render callbacks are
  unsupported, widgets/`useWidget` are deferred, context is partial, and
  interleaved views retain existing constraints;
- development-only warning behavior and production silence.

Add the new guide to the existing React Fiber docs sidebar, update the concise
API/get-started pages to point to it without displacing the native `<layer>` /
`<view>` guidance, and add a narrowly worded `docs/whats-new.md` release note.

**Upstream requirement:** `proposal.md` documentation section; `design.md`
Decisions 7–9; `tasks.md` §6.1, §6.3, and §6.4.

**Compatibility/release consideration:** No new example application or
unrelated existing example rewrite is required. If a release changes the
published package surface, follow this repository's tag-based publishing flow
rather than creating an upstream-style changeset. Release text must not claim
full `@deck.gl/react` parity or widget support.

## 4. Validation Plan

Run validation after the implementation work, not as part of this planning task.

1. **Package tests:** run
   `yarn workspace @deck.gl-community/react-fiber test`.
   This executes the current Vitest/jsdom project plus the new compat tests.
   It validates the upstream component, wrapper, view-composition, alias,
   ref-inventory, context, warning, and negative-scope scenarios.

2. **Targeted export/declaration verification:** build the workspace package
   through the repository build process, then verify that all five published
   compat specifiers resolve in both ESM and CommonJS and each has a matching
   `.d.ts` file under `modules/react-fiber/dist/compat/`. Assert that the root
   package/default DOM exports continue to resolve unchanged. This directly
   validates the source-to-target export-map adaptation.

3. **Type validation:** run `yarn build` (the repository's declaration/type and
   package build gate) and ensure the new compat type tests compile. Add a
   consumer-shaped type fixture/import check for `DeckGLProps`, `DeckGLRef`,
   `DeckGLContextValue`, all wrapper matrices, aliases, and the intentional
   absence of low-level props. Confirm no generated declaration imports
   `@deck.gl/react`.

4. **Lint and formatting:** run `yarn lint`, followed by `yarn lint-fix` only
   for changed files and review the resulting diff. This follows the repository
   Biome/Ocular gate without reformatting unrelated files.

5. **Documentation site:** run `yarn --cwd website build`. It must resolve the
   new migration-guide sidebar entry, links, and code examples without broken
   references. This covers the upstream documentation-consistency requirement.

6. **Manual migration smoke tests:** using the existing workspaces—not a new
   app—validate the following in `examples/react-fiber/vite` (and one
   SSR/client-boundary example such as `examples/react-fiber/nextjs`):
   - change a representative `DeckGL` import to the compat root and import a
     `ScatterplotLayer` wrapper from `compat/layers`;
   - render a stable-ID layer and an explicit-ID `MapView` with nested compat
     layers in standard mode;
   - use a ref to call each supported picking method only after initialization;
   - supply a custom `ContextProvider` and observe its `deck` field;
   - confirm a documented unsupported child/prop warns only in development;
   - retain the existing interleaved map example without adding a compat view
     and confirm its known limitation is documented rather than hidden.

7. **Regression checks:** retain the existing DOM/reconciler tests for
   standalone canvas mode, explicit canvas/parent behavior, interleaved roots,
   root cleanup, native `<layer>`/`<view>` creation, and renderer store
   isolation. Compat must not regress these existing contracts while adapting
   the official React surface.

## 5. Scope Boundaries and Risks

### Deliberately excluded work

- No new workspace/package and no restoration of upstream's four-package
  monorepo layout.
- No modification of the existing native `Deckgl`, reconciler, store, or JSX
  intrinsic APIs unless a compat test proves a narrowly necessary integration
  change.
- No full `@deck.gl/react` prop, context, component, ref, or behavior parity.
- No support for widgets, JSX widget wrappers, `useWidget`, or widget context.
- No support for official function-child render callbacks.
- No wrappers outside the explicit 32-layer/five-view matrices, including
  aggregation internals/helper/type exports.
- No unrelated dependency upgrades, renderer refactors, documentation cleanup,
  or changes to existing example applications.
- No upstream-style Changesets; use the repository's established publishing
  process when releasing the public API.

### Adaptation risks and required decisions

1. **Upstream artifact conflicts:** Six upstream specs retain merge markers;
   the obsolete sides promise broad props/context/widget behavior and conflict
   with the proposal/design/task scope. The implementation owner must record
   the narrowed scope as the accepted interpretation before coding. The
   standalone `compat-widget-hook/spec.md` is contradictory and remains out of
   scope.

2. **Unspecified pre-mount ref behavior:** The accepted change fixes the ref
   inventory and types `deck` as optional, but does not say whether picking
   methods throw, no-op, or otherwise behave before deck initialization. Decide
   and document this behavior before implementing ref forwarding; do not adopt
   the conflict-marker-only behavior by assumption.

3. **Unspecified layer-wrapper child/ref semantics:** The accepted source
   requires view-to-layer nesting but does not define whether individual layer
   wrappers forward `children`, expose refs, or promise constructor-prop parity.
   Keep the initial wrapper contract minimal. If forwarding layer children is
   needed for a migration case, add an explicit API decision and test rather
   than silently treating it as full official behavior.

4. **Export-map completeness:** This package publishes conditional ESM, CJS,
   and declaration targets. A source file alone is insufficient; each compat
   subpath needs a generated artifact and a matching `package.json` entry.
   Missing any one variant breaks a class of consumer.

5. **Consolidated shared state:** The existing renderer now has per-root stores,
   but `useDeckgl()` is backed by a legacy shared store. The compat ref/context
   bridge can expose the wrong deck when multiple roots mount or unmount.
   Multi-root lifecycle coverage is required before public release.

6. **Interleaved mode:** The current reconciler deliberately does not set
   `views` on `MapboxOverlay`. Compat view wrappers must not mask or circumvent
   that limitation; documentation and tests must make it visible.

7. **Partial context claims:** Passing a caller's `ContextProvider` must not
   make consumers believe the official fields update correctly. Only `deck` is
   contractual until independently implemented and validated.

8. **React/SSR boundaries:** `src/dom/components.tsx` imports `client-only`.
   Compat root exports must preserve the current client-only behavior and be
   checked in an existing Next.js consumer boundary; they must not introduce a
   server-renderable substitute.

```

docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md
```diff
@@ -0,0 +1,140 @@
...
+          getPosition={point => point.position}
+          getRadius={100}
+        />
+      </MapView>
+    </DeckGL>
+  );
+}
+```
+
+The native `Deckgl`, `<layer>`, and `<view>` APIs remain available from the
```
chrisgervang: I can't help but feel like we should try to minimize the differences between the upstream API and fiber API, when it's purely cosmetic.

The migration diffs can be smaller if the native component was also spelled as `DeckGL`

I'm looking at this from the lens of minimizing breaking changes and encouraging use of the native API over time

docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md
```diff
@@ -0,0 +1,140 @@
...
+
+- `gl`, `canvas`, `parent`, and `_customRender`: compat does not accept these
+  low-level renderer ownership props. Continue using the native `Deckgl`
+  component when they are required.
+- Function children: compat does not support function-child render callbacks.
+  Render supported layers and views directly instead.
+- Widgets, JSX widget wrappers, and `useWidget`: these APIs are deferred and
+  unavailable from compat.
+- Extra layer or view components: only the export matrix above is supported.
+- Interleaved rendering: `MapboxOverlay` owns its views. Compat view wrappers
```
chrisgervang, line 136: We also just released a `MaplibreOverlay` in 9.4 that adds v6 support. The APIs and rendering technologies are diverging so it was a good time to split.

Deck v10 will drop Maplibre support from the "Mapbox" module too so that the internals can be simplified, as well

modules/react-fiber/src/compat/deckgl.tsx
```diff
@@ -0,0 +1,122 @@
...
+    );
+  }
+}
+
+/**
+ * A bounded migration adapter for supported `@deck.gl/react` applications.
+ *
+ * It lowers compatible layer and view components to the local native reconciler.
+ */
+export const DeckGL = forwardRef<DeckGLRef, DeckGLProps>(function DeckGL(props, ref) {
```
chrisgervang, line 77: This is a great place to start. Thoughts on converging the native component to this API?

The compatibility layer seems like it could be entirely focused on how children elements are written

modules/react-fiber/src/compat/geo-layers.tsx
```diff
@@ -0,0 +1,80 @@
...
+);
+
+/** @deprecated Use MVTLayer. */
+export const mVTLayer = MVTLayer;
+/** @deprecated Use MVTLayer. */
+export const mvtLayer = MVTLayer;
+/** @deprecated Use WMSLayer. */
+export const wMSLayer = WMSLayer;
+/** @deprecated Use WMSLayer. */
+export const wmsLayer = WMSLayer;
```
chrisgervang, line 80: Out of curiosity, why do we have spelling aliases for these layers and not others?

My understanding is deck always styled acronyms in its API with capitalization

modules/react-fiber/src/compat/types.ts
```diff
@@ -0,0 +1,48 @@
+import type {Deck} from '@deck.gl/core';
+import type {MapboxOverlay} from '@deck.gl/mapbox';
+import type {ReactNode} from 'react';
+import type {DeckglProps} from '../types/index';
+
+/** The deck.gl instance exposed by the compatibility adapter. */
+export type DeckGLInstance = Deck | MapboxOverlay;
```
chrisgervang, line 7: It seems like there is an open question on how to support the ~3 officially supported basemap overlays, plus the additional semi-official bindings.

The final solution will likely need to preserve bundle sizes, and ideally extensibility so that interface conforming non-official bindings may register themselves

chrisgervang (approved):
Fantastic start! Thank you for jumping into this with us.

I left some follow ups notes for discussion - overall, my goal is to convert as many react users over to this as possible in the v10 release as there are many many benefits to this approach.

I believe holding a principle of path of least resistance, and reducing choices will increase adoptions.

Where ever possible I'm hoping we can minimize breaking changes and differences between _current_ deck v9, the native fiber, and the compatibility fiber API surfaces.

And, I want to eventually challenge the need for a compatibility layer - if it causes more confusion than helps reduce migration cost.

Part of this a question of how to reduce confusion for LLMs, which historically fail to learn new APIs. For manually authored code, we could provide a codemod.

modules/react-fiber/src/compat/geo-layers.tsx
brandonjpierce: Purely an aesthetics choice, since a custom react host component (e.g. think `<div>`, `<button>`, `<input>`, etc...) cannot be PascalCase I just thought `<wMSLayer />` looked gross 😆. Thus we aliased to allow for the more pleasing `<wmsLayer />`. This is not a concern now that we moved to the lower level `<layer />` primitive.

docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md
brandonjpierce: Yeah I think that is completely sensible. The refactor lift in theory should be quite low for folks already adopting the fiber renderer since the number of `<Deckgl />` components in an app should be near 1. 

This is a breaking change semver wise from the original package but since we are already changing the version number for the deckgl-community package I don't think this is that large of an issue.

modules/react-fiber/src/compat/deckgl.tsx
brandonjpierce: Yes will make this change per convo above 👍

modules/react-fiber/src/compat/types.ts
brandonjpierce: Yeah this always felt a bit awful in the code but I couldn't find a great feeling alternative for this

```
// modules/react-fiber/src/reconciler/renderer.ts

const isOverlay = 'interleaved' in props;
const deckgl = isOverlay ? new MapboxOverlay(props) : new Deck(props);
```

This bit of code (and some tangential things surrounding it) all get a bit more complex with the inclusion of both `MapboxOverlay` and `MaplibreOverlay` and how that impacts the sort of available props at the tail end from a consumers perspective. 

One thing I purposefully do NOT do in this bit of code (for some reasons I can explain if needed) is return `._deck` from the `new MapboxOverlay` instance to have it match the return structure of `new Deck`. In some cases we found it valuable to be able to also look at the Mapbox/Maplibre instance from our `useDeckgl` hook especially in situations where we were in interleaved mode.

Open to ideas on how to smooth this portion out a bit and perhaps this is a broader play into future `react-map-gl` updates.

docs/modules/react-fiber/developer-guide/migrate-from-deckgl-react.md
brandonjpierce: Resolved with https://github.com/visgl/deck.gl-community/pull/761/changes/85ba40574f43e3d9b852a5a6ee0ef394f61483e1

modules/react-fiber/src/compat/deckgl.tsx
brandonjpierce: Resolved with https://github.com/visgl/deck.gl-community/pull/761/changes/85ba40574f43e3d9b852a5a6ee0ef394f61483e1

modules/react-fiber/src/compat/types.ts
brandonjpierce: This does tie nicely into an open issue from `@am2222`:

https://github.com/deckgl-fiber-renderer/fiber.gl/issues/28

modules/react-fiber/src/compat/geo-layers.tsx
chrisgervang: Gotcha, yeah I agree - the `<layer/>` version is a nice primitive.

Could we drop the extra compatibility aliases to avoid redundancy? I believe the upstream deck code only supports PascalCase with capital acronyms

modules/react-fiber/src/compat/types.ts
chrisgervang: I quite like the separate entry point style

> Separate entry point — @deckgl-fiber-renderer/dom/maplibre (or a maplibre export condition) that binds MapLibreOverlay, keeping the default import Mapbox-only.

`react-map-gl` has adopted this pattern, and it's worked out nicely. package.json supports in a standard way now too, which can be tree shaken.

It pulls the branching up into the application level, which means users need to be explicit.

The benefit of this means we can customize the typing of each entry point, and provide a generic entry point with relaxed typing for custom overlays

Wdyt?

chrisgervang: > perhaps this is a broader play into future `react-map-gl` updates.

I could imagine a future where we have a super smooth interleaved experience where deck and basemap are managed in the same fiber, where basemap "layer" and deck "layer" are routed seamlessly.

But I don't want to get too ahead of myself.. moving each into their own fiber is the logical first phase 😄

brandonjpierce: For the sake of cleaner PRs that are easier to review for the team, would you mind if we kept the scope of this PR to the backwards compat API and I can do a separate PR addressing this scope of work/change?

modules/react-fiber/src/compat/geo-layers.tsx
brandonjpierce: Aliases have been removed 👍

chrisgervang: I still see a couple like `wMSLayer`

brandonjpierce: Per my earlier note, you have to use `camelCase` or all lowercase for react host components. PascalCase gives a hint to React that it is a custom component (a function).

https://react.dev/reference/react/createElement#caveats

> When you use JSX, you must start a tag with a capital letter to render your own custom component. In other words, <Something /> is equivalent to createElement(Something), but <something /> (lowercase) is equivalent to createElement('something') (note it’s a string, so it will be treated as a built-in HTML tag).

So the legacy fiber jsx tag is `<wMSLayer />`, the new fiber API is `<layer layer={new WmsLayer()} />` and the backwards compat export (what we did in this PR) is `<WMSLayer />`

The alias I removed was for the legacy fiber jsx tag to support `<wmsLayer />` (which is now marked as deprecated alongside `<wMSLayer />`).

chrisgervang: Why does this repo need the legacy fiber JSX tag? The new fiber API and deck compat versions seem to cover the necessary cases, and ideally we can keep it to two variations

brandonjpierce: It was mostly a backwards compatibility thing so folks could transition to the new `<layer />` api iteratively while not breaking existing apps that used the legacy fiber api e.g. `<scatterplotLayer />`.

However, given this packages new home I think it would be fair to start in a more "clean" state and fully remove the legacy jsx api if you would prefer. That cleans quite a few things up in the code.

chrisgervang: Yeah, let's take the opportunity to clean up. If you want to split cleanup into follow up PRs, that's fine with me.

(The alias deletion seems like a good one to keep in this PR)

brandonjpierce: Sounds good! I'll put the cleanup work in this PR since it seems reasonable to bundle up with the @deck.gl/react backwards compat changes. I'll reserve https://github.com/visgl/deck.gl-community/pull/761#discussion_r4050550782 for a follow on.

brandonjpierce: Legacy API removed in 1db8fc01e38d464801606030f186990a2203f2f8

brandonjpierce:
> Part of this a question of how to reduce confusion for LLMs, which historically fail to learn new APIs. For manually authored code, we could provide a codemod.

Codemod would be a great idea! We also have someone working on an agent skill already here: https://github.com/gohypergiant/agent-skills/pull/156

I think the codemod + agent skill + a llms.txt file should give us a good result with agent usage. I'll add those to the top level fiber issue tracker.

docs/modules/react-fiber/api-reference/deckgl.md
```diff
@@ -1,19 +1,52 @@
...
+`onDeckglChange` receives the `Deck` or `MapboxOverlay` instance owned by this
+`DeckGL` root after configuration. It receives `null` when the root cleans up.
+Use local state when another component needs the instance:
+
+```tsx
+import {useEffect, useState} from 'react';
+import {DeckGL, type DeckglInstance} from '@deck.gl-community/react-fiber';
+
+export function Map() {
+  const [deckgl, setDeckgl] = useState<DeckglInstance | null>(null);
```
chrisgervang, line 39: I see this block of code is documented in full a couple of times. Does it make sense to do that?

chrisgervang (approved):
Nice cleanup! So, custom layers are straightforward now.. just use `<layer>`? As opposed to extending/registering.

Also, can users mix the compat syntax and the new syntax? Just thinking out loud about how the transition would be branded. It'd be nice if it could support a mix for incremental adoption.

brandonjpierce:
> Also, can users mix the compat syntax and the new syntax? Just thinking out loud about how the transition would be branded. It'd be nice if it could support a mix for incremental adoption.

Yes, I am updating the docs now but something like this is completely valid:


```markdown
Compat components turn into the same native `<layer>` and `<view>` elements
used by React Fiber. You can use both styles in the same `DeckGL` tree. This
lets you move to compat one layer at a time:
```

```tsx
import {ScatterplotLayer} from '@deck.gl/layers';
import {DeckGL} from '@deck.gl-community/react-fiber/compat';
import {PolygonLayer} from '@deck.gl-community/react-fiber/compat/layers';

export function MixedLayers() {
  return (
    <DeckGL initialViewState={{longitude: -122.4, latitude: 37.8, zoom: 12}}>
      <layer layer={new ScatterplotLayer({id: 'native-points', data: []})} />
      <PolygonLayer id="compat-polygons" data={[]} />
    </DeckGL>
  );
}
```

---

Edit: updated in https://github.com/visgl/deck.gl-community/pull/761/commits/f805cbaf4107f0db21b894e4a5296c7f4647fb4c and https://github.com/visgl/deck.gl-community/pull/761/commits/8c98608150ce86fc861e02f48d38e6819dbc93c2

chrisgervang (approved):

docs/modules/react-fiber/api-reference/deckgl.md
brandonjpierce: Adding a todo to clean up / expand examples https://github.com/visgl/deck.gl-community/issues/747#issuecomment-5762108290

